PR #2, opened 2018-01-12
Initial support for http/2(h2, h2c) and gRPC clients/upstream severs

### LINKED ISSUE #1: http/2(h2, h2c) and gRPC support
@brancz Hi, thanks for sharing a very inspiring project 👍 

I'm POCing to add support for gRPC but I'm not really sure if it is really feasible.
I'd greatly appreciate if you could leave comments on the state of my POC.
Thanks!

https://github.com/mumoshu/kube-rbac-proxy/tree/grpc-support

In nutshell:

- It runs a [gRPC proxy instead of the http(s) one](https://github.com/mumoshu/kube-rbac-proxy/blob/grpc-support/main.go#L169-L178) when a specific flag is passed.
- In the notion of gRPC we don't have request path/resource name but do have "method name" instead.
  - Given that, my idea is that I could take a gRPC method name as a `nonResourceURL` in K8S' RBAC policy, so that we could authorize the method call w/ RBAC
- I'm relying on a bearer token passed via the `authorization` header(=`metadata` in gRPC?) to be used for authentication

My end-goal/intended use-case of adding a gRPC support to kube-rbac-proxy is achieve authn/authz for Helm/Tiller. I don't like to implemented a yet another RBAC system inside Helm/Tiller but rather reuse K8S RBAC instead.

https://github.com/kubernetes/helm/issues/1918#issuecomment-356548715

Ideally, kube-rbac-proxy could authenticate the client and authorize the rpc call w/ RBAC. Once authorized, rbac-proxy could add the authn result to metadata to be used by tiller(upstream of kube-rbac-proxy) to "impersonate" the user. Tiller could then CRUD k8s resources as the user.

WDYT?

mumoshu:
Resolves #1

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -84,14 +84,16 @@ type kubeRBACProxyAuth struct {
 	authorizer.RequestAttributesGetter
 	// authorizer determines whether a given authorization.Attributes is allowed
 	authorizer.Authorizer
+	// config for kube-rbac-proxy
+	Config AuthConfig
 }
 
-func newKubeRBACProxyAuth(authenticator authenticator.Request, authorizer authorizer.Authorizer, authzConfig *AuthzConfig) AuthInterface {
-	return &kubeRBACProxyAuth{authenticator, newKubeRBACProxyAuthorizerAttributesGetter(authzConfig), authorizer}
+func newKubeRBACProxyAuth(authenticator authenticator.Request, authorizer authorizer.Authorizer, authConfig AuthConfig) AuthInterface {
+	return &kubeRBACProxyAuth{authenticator, newKubeRBACProxyAuthorizerAttributesGetter(authConfig.Authorization), authorizer, authConfig}
 }
 
 // BuildAuthHandler creates an authenticator, an authorizer, and a matching authorizer attributes getter compatible with the kube-rbac-proxy
-func BuildAuthHandler(client clientset.Interface, config AuthConfig) (*Handler, error) {
+func BuildAuthHandler(client clientset.Interface, config AuthConfig) (AuthInterface, error) {
 	// Get clients, if provided
 	var (
 		tokenClient authenticationclient.TokenReviewInterface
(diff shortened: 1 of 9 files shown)


mumoshu:
@brancz Would you like me to rebase this onto the master branch?

brancz:
yes, rebasing would be great, thanks!

mumoshu:
@brancz Done 👍

main.go
```diff
@@ -157,6 +169,55 @@ func main() {
...
+		//
+		// See https://github.com/golang/go/issues/14141 for more context.
+		//
+		// Possible alternative:
+		//
+		// We could potentially use grpc-go server's HTTP handler support
+		// which would handle HTTP UPGRADE from http1.1 to http/2, especially in case
+		// what you wanted kube-rbac-proxy to authn/authz was gRPC over h2c calls.
+		//
+		// // Note that golang's http server requires a client(including gRPC) to send HTTP Upgrade req to
```
brancz: double `//`

mumoshu:
Thanks for your review @brancz!
Every point sounded great, and now I believe I've addressed all the comments 👍

auth.go
```diff
@@ -204,12 +205,11 @@ func (n krpAuthorizerAttributesGetter) GetRequestAttributes(u user.Info, r *http
 	return attrs
 }
 
-type Handler struct {
-	AuthInterface
-	Config AuthConfig
+type Handler interface {
```
brancz: This would be an `AuthHandler` as it returns whether authN and authZ was successful.

auth.go
```diff
@@ -204,12 +205,11 @@ func (n krpAuthorizerAttributesGetter) GetRequestAttributes(u user.Info, r *http
...
+	Handle(w http.ResponseWriter, req *http.Request) bool
 }
 
-func (h *Handler) Handle(w http.ResponseWriter, req *http.Request) bool {
+func (h *kubeRBACProxyAuth) Handle(w http.ResponseWriter, req *http.Request) bool {
```
brancz: Doc string missing.

mumoshu:
@brancz Thanks again! I've addressed the two comments just now.

brancz:
Awesome, thanks! lgtm :+1: 

(if you're interested it would be good to add some testing :slightly_smiling_face:, happy to discuss on Kubernetes slack if you're interested, I'm brancz there as well)

mumoshu:
@brancz Yes, I'm willing to write some tests!
In the meantime, please let me share you how I've tested this in combination with my fork of helm/tiller: https://github.com/mumoshu/helm/commit/58b13bac4109139b2437ae2939ac56e660b96cae
It would allow anyone interested to test the combo manually anyway!

brancz:
Nice! Yes I'll come up with some first tests for the other existing functionality, then we can see how we can add tests for this functionality.

PR #5, opened 2018-03-27
Make TLS version configurable

jpiper:
I haven't tested this, but it should work...

I've set the default to TLS 1.2, which aligns with the default k8s minimum TLS version.

### DIFF
--- a/main.go
+++ b/main.go
@@ -19,6 +19,7 @@ package main
 import (
 	"crypto/tls"
 	stdflag "flag"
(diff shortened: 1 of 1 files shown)


jpiper:
I've tested this patch on v0.2.0 - it doesn't seem to work on HEAD, but that's because HEAD itself is broken for me right now

brancz:
lgtm :+1: thanks!

PR #7, opened 2018-03-27
Build containerised app in docker

jpiper:
More reproducable than cross-compiling on your local box and injecting into a container


brancz:
lgtm, thanks!

PR #13, opened 2018-08-09
Add OIDC support for kube-rbac-proxy

piotrmsc:
This PR adds ID token validation to the kube-rbac-proxy. 
If OIDC configuration is provided proxy validates token based on provided config, otherwise delegates token to apiserver. 
Project structure was slightly changed.


s-urbaniak:
@brancz pinged me to help out with this PR. First, thanks a lot for the contribution! OIDC looks like a great addition. To get started, do you mind to split the vendoring part and the actual changes in two separate commits?

piotrmsc:
@s-urbaniak Hi, sure, sorry for late response...
I will push the changes only in the code and glide.yaml to allow local glide install.
I'm also open for discussion about this PR ie. on slack ;-) 

PS. Do you plan to use dep instead of glide for dep. mgmt. ?

s-urbaniak:
@piotrmsc great, thank you! I don't know of plans to move to dep right now, I guess this could happen at some point, but @brancz should have more insights.

Are you on the k8s slack?

brancz:
I'm fine with using dep instead of glide, no strong opinions.

piotrmsc:
Changes pushed and yes, I'm on k8s slack, I will ping you there.
Regarding my dep question, I was just curious, I don't want to 'destroy everything' ;-)

main.go
```diff
@@ -148,6 +182,27 @@ func main() {
...
+		//If proxy is set to kubernetes service(apiserver) add SA token to the request
+		if strings.Contains(cfg.upstream, "kubernetes.default") {
+			director := proxy.Director
+			proxy.Director = func(req *http.Request) {
+				director(req)
+				req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", kcfg.BearerToken))
+			}
+		}
+
+		if cfg.upstreamInsecure {
```
s-urbaniak: This change seems to be unrelated to the actual OIDC contribution as it adds the possibility to skip TLS verification for upstream. Two comments on this:

1. This `if` statement configures the proxy for every request which seems unnecessary and potentially racy. Do you mind to factor this somewhere around line 250?

2. The proxy transport gets also overwritten if h2c is enabled, so `upstreamInsecure` and `upstreamForceH2C` are mutually exclusive. We have to take care to initialize the http transport if `upstreamInsecure=true` _before_ the h2c 
check in line 250.

main.go
```diff
@@ -148,6 +182,27 @@ func main() {
 			return
 		}
 
+		//If proxy is set to kubernetes service(apiserver) add SA token to the request
+		if strings.Contains(cfg.upstream, "kubernetes.default") {
```
s-urbaniak: This seems very hardcoded to me and I don't seem to understand the use case completely. Do I understand it correctly that you are putting an RBAC proxy before the API server? I am unsure we want to support this use case here as the API server itself takes care of authn/authz already.

pkg/authn/oidc.go
```diff
@@ -0,0 +1,40 @@
...
+	CAFile               string
+	UsernameClaim        string
+	UsernamePrefix       string
+	GroupsClaim          string
+	GroupsPrefix         string
+	SupportedSigningAlgs []string
+}
+
+// NewOIDCAuthenticator returns OIDC authenticator 
+func NewOIDCAuthenticator(authn *AuthnConfig) (authenticator.Request, error) {
```
s-urbaniak: nit: just pass a `*OIDCConfig` here, we don't need the whole dependency on `*AuthnConfig`.

main.go
```diff
@@ -258,3 +313,19 @@ func main() {
 		glog.Info("Received SIGTERM, exiting gracefully...")
 	}
 }
+
+// Returns intiliazed config, allows local usage (outside cluster) based on host kubeconfig or in-cluter
+func initKubeConfig() *rest.Config {
+
+	kubeConfigLocation := filepath.Join(os.Getenv("HOME"), ".kube", "config")
```
s-urbaniak: this location is very hardcoded, do you mind to add a dedicated `kubeconfig` location? If unset we could fallback to the in-cluster-config.

examples/oidc/README.md
```diff
@@ -0,0 +1,119 @@
+# Secure a service with OIDC authentication
+
+> Note to try this out with minikube, make sure you enable RBAC correctly as explained [here](../minikube-rbac).
+
+This example ilustrates how to use `kube-rbac-proxy` with OIDC support to secure k8s apiserver.
```
s-urbaniak: ahh, ... now once I reached this sentence I recognize that securing the API server is indeed your canonical use case here ;-)

also pinging @brancz: is securing the API server itself, using the kube-rbac-proxy, really something that is in scope for this project? I believe, it is not, but I may be mistaken here.

Independent, it would be great to have an example, similar to i.e. the `non-resource-url` one, where an arbitrary Pod is being secured using this proxy.

main.go
```diff
@@ -106,36 +115,61 @@ func main() {
 	flagset.StringVar(&cfg.auth.Authentication.Header.UserFieldName, "auth-header-user-field-name", "x-remote-user", "The name of the field inside a http(2) request header to tell the upstream server about the user's name")
 	flagset.StringVar(&cfg.auth.Authentication.Header.GroupsFieldName, "auth-header-groups-field-name", "x-remote-groups", "The name of the field inside a http(2) request header to tell the upstream server about the user's groups")
 	flagset.StringVar(&cfg.auth.Authentication.Header.GroupSeparator, "auth-header-groups-field-separator", "|", "The separator string used for concatenating multiple group names in a groups header field's value")
+
+	//Authn OIDC flags
+	flagset.StringVar(&cfg.auth.Authentication.OIDC.IssuerURL, "oidc-issuer", "", "Issuer URL")
```
s-urbaniak: nit: let's clarify, that only HTTPS schemes will be accepted. Also being optional here, because if unset, no oidc will be used.

main.go
```diff
@@ -106,36 +115,61 @@ func main() {
...
+	flagset.StringVar(&cfg.auth.Authentication.OIDC.ClientID, "oidc-clientID", "", "Registered clientID at Identity Provider")
```
s-urbaniak: let's clarify, similar to upstream, that this must be set, if the issuer URL is set, see https://github.com/kubernetes/kubernetes/blob/v1.11.2/pkg/kubeapiserver/options/authentication.go#L198-L199

main.go
```diff
@@ -106,36 +115,61 @@ func main() {
...
 	kubeClient, err := kubernetes.NewForConfig(kcfg)
 	if err != nil {
 		glog.Fatalf("Failed to instantiate Kubernetes client: %v", err)
 	}
 
-	if cfg.resourceAttributesFile != "" {
-		b, err := ioutil.ReadFile(cfg.resourceAttributesFile)
+	var authenticator authenticator.Request
+	// If OIDC configuration provided, use oidc authenticator
+	if &cfg.auth.Authentication.OIDC != nil {
```
s-urbaniak: The flags framework will always set `cfg.auth.Authentication.OIDC` to be non-nil. I suggest to add this check instead:
```
if cfg.auth.Authentication.OIDC.IssuerURL != ""
```

Otherwise, by starting the proxy it will panic like this:
```
$ kube-rbac-proxy 
F0829 12:30:28.325892   23957 main.go:146] Failed to instantiate OIDC authenticator: 'oidc-issuer-url' ("") has invalid scheme (""), require 'https'
goroutine 1 [running]:
github.com/brancz/kube-rbac-proxy/vendor/github.com/golang/glog.stacks(0xc420264100, 0xc420066000, 0x95, 0xe9)
	/home/sur/go/src/github.com/brancz/kube-rbac-proxy/vendor/github.com/golang/glog/glog.go:769 +0xcf
...
```

examples/oidc/README.md
```diff
@@ -0,0 +1,119 @@
+# Secure a service with OIDC authentication
```
s-urbaniak: We should maybe also point out more clearly, that this is explicitly tailored for the use case when kubernetes is *not* configured to use OIDC for authentication.

If kubernetes *is* configured with OIDC, one should still use the existing delegating authn provider (which passes the OIDC ID token to the token review).

main.go
piotrmsc: The motivation for this was a feature allowing to secure api-server in the environments where admins cannot modify api-server configurations like OIDC. But I do accept that this is not desired feature and I will remove it

main.go
piotrmsc: Yes, you're right, skipping tls for the upstream will be removed

README.md
```diff
@@ -47,6 +48,13 @@ Usage of _output/linux/amd64/kube-rbac-proxy:
...
   -v, --v Level                                     log level for V logs
       --vmodule moduleSpec                          comma-separated list of pattern=N settings for file-filtered logging
+      --oidc-issuer                                 The URL of the OpenID issuer, only HTTPS scheme will be accepted. If                                               set, it will be used to verify the OIDC JSON Web Token (JWT).
+      --oidc-clientID                               The client ID for the OpenID Connect client, must be set if                                                        oidc-issuer-url is set.
+      --oidc-groups-claim                           Identifier of groups in JWT claim, by default set to 'groups'
+      --oidc-username-claim                         Identifier of the user in JWT claim, by default set to 'email'
+      --oidc-groups-prefix                          If provided, all groups will be prefixed with this value to prevent                                                conflicts with other authentication strategies.
+      --oidc-sing-alg                               Supported signing algorithms, default RS256
```
s-urbaniak: typo nit: s/--oidc-sing-alg/--oidc-sign-alg

examples/oidc/README.md
```diff
@@ -0,0 +1,24 @@
+# Secure a service with OIDC authentication in a non-OIDC sercured Kubernetes cluster
```
s-urbaniak: Let's set a short and tidy header, i.e. `# OIDC authentication` and explain the rationale in a first paragraph. Suggestion:

> Kubernetes supports OIDC authentication natively, configured via API server flags. In this case, a request contains the OIDC ID token in its bearer header. The token is then used for token review by the Kubernetes API. This already works with the existing kube-rbac-proxy functionality.  The resulting metadata (user and/or group information) is then passed to the subject review Kubernetes API for authorization.

> If Kubernetes is not configured to use OIDC and changes to the API server are not possible (i.e. in 3rd party or restricted environments), kube-rbac-proxy can be configured to authenticate the request against OIDC itself. In this case, the token review functionality of Kubernetes is omitted. As above the resulting metadata is still passed to the subject review Kubernetes API for authorization.

/cc @brancz

examples/oidc/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+> Note to try this out with minikube, make sure you enable RBAC correctly as explained [here](../minikube-rbac/README.md)
+
+This example ilustrates how to use `kube-rbac-proxy` with OIDC support to secure sample pod using provided OIDC configuration.
+
+Requests after sucessful authentication and authorization will be proxied to configured pod. When OIDC configuration is specified, proxy does not delegate sent token to the apiserver using TokenReviews, instead,validates sent JWT itself based on provided configuration.
```
s-urbaniak: let's move this to the first paragraph and merge with the suggestion above.

examples/oidc/README.md
```diff
@@ -0,0 +1,24 @@
...
+```
+```bash
+$ kubectl create -f configmap.yaml
+```
+
+```bash
+$ kubectl create -f client.yaml
+```
+
+> Note you have to replace {ISSUER} and {CLIENT_ID} in the deployment with issuer and client from your OIDC provider configuration.
```
s-urbaniak: Let's remove the leading `> ` as it is not a quote.

s-urbaniak: Language nit:
> Note: The {ISSUER} and {CLIENT_ID} in the deployment have to be replaced with the issuer and client in the OIDC provider configuration.

examples/oidc/README.md
```diff
@@ -0,0 +1,24 @@
...
+$ kubectl create -f configmap.yaml
+```
+
+```bash
+$ kubectl create -f client.yaml
+```
+
+> Note you have to replace {ISSUER} and {CLIENT_ID} in the deployment with issuer and client from your OIDC provider configuration. 
+
+For an identity provider you can use ie. [dex](https://github.com/dexidp/dex)
```
s-urbaniak: I would omit this (, although I really really like Dex) :-D

examples/oidc/deployment.yaml
```diff
@@ -0,0 +1,87 @@
...
+        app: kube-rbac-proxy
+    spec:
+      # {BEGIN} for minikube development only if OIDC provider is deployed in minikube itself ie. dex
+      hostNetwork: true
+      dnsPolicy: ClusterFirstWithHostNet
+      # {END} 
+      serviceAccountName: kube-rbac-proxy
+      containers:
+      - name: kube-rbac-proxy
+        image: docker.io/piotrmsc/kube-rbac-proxy:0.3.2
```
s-urbaniak: Note to ourselfes: we should change this `docker.io/piotrmsc` once merged.

main.go
```diff
@@ -106,38 +111,66 @@ func main() {
...
+
+	//Authn OIDC flags
+	flagset.StringVar(&cfg.auth.Authentication.OIDC.IssuerURL, "oidc-issuer", "", "The URL of the OpenID issuer, only HTTPS scheme will be accepted. If set, it will be used to verify the OIDC JSON Web Token (JWT).")
+	flagset.StringVar(&cfg.auth.Authentication.OIDC.ClientID, "oidc-clientID", "", "The client ID for the OpenID Connect client, must be set if oidc-issuer-url is set.")
+	flagset.StringVar(&cfg.auth.Authentication.OIDC.GroupsClaim, "oidc-groups-claim", "groups", "Identifier of groups in JWT claim, by default set to 'groups'")
+	flagset.StringVar(&cfg.auth.Authentication.OIDC.UsernameClaim, "oidc-username-claim", "email", "Identifier of the user in JWT claim, by default set to 'email'")
+	flagset.StringVar(&cfg.auth.Authentication.OIDC.GroupsPrefix, "oidc-groups-prefix", "", "If provided, all groups will be prefixed with this value to prevent conflicts with other authentication strategies.")
+	flagset.StringArrayVar(&cfg.auth.Authentication.OIDC.SupportedSigningAlgs, "oidc-sing-alg", []string{"RS256"}, "Supported signing algorithms, default RS256")
```
s-urbaniak: typo: `oidc-sign-alg`

main.go
```diff
@@ -106,38 +111,66 @@ func main() {
...
+	flagset.StringVar(&cfg.auth.Authentication.OIDC.CAFile, "oidc-ca-file", "", "If set, the OpenID server's certificate will be verified by one of the authorities in the oidc-ca-file, otherwise the host's root CA set will be used.")
+
+	//Kubeconfig flag
+	flagset.StringVar(&cfg.kubeconfigLocation, "kubeconfig", "", "Specifies location of the kubeconfig on the host machine. If not set in cluster configuration will be used")
```
s-urbaniak: Description nit:
`"Path to a kubeconfig file, specifying how to connect to the API server. If unset, in-cluster configuration will be used".

main.go
```diff
@@ -258,3 +291,22 @@ func main() {
...
 	}
 }
+
+// Returns intiliazed config, allows local usage (outside cluster) based on provided kubeconfig or in-cluter
+func initKubeConfig(kcLocation string) *rest.Config {
+
+	if kcLocation != "" {
+		kubeConfig, err := clientcmd.BuildConfigFromFlags("", kcLocation)
+		if err != nil {
+			glog.Fatal("unable to build rest config based on provided kubeconfig location. Trying in-cluster configuration")
```
s-urbaniak: This error message contradicts actual behavior. If loading kubeconfig fails, we do `glog.Fatal` here, effectively `os.Exit(255)`ing, see [1]. We should either change the description or change the behavior.

[1] https://godoc.org/github.com/golang/glog#Fatal

pkg/authn/config.go
```diff
@@ -0,0 +1,24 @@
+package authn
+
+type AuthnHeaderConfig struct {
```
s-urbaniak: Let's add some top-level godoc to this struct.

pkg/authn/config.go
```diff
@@ -0,0 +1,24 @@
...
+	// to tell the upstream server about the user's name
+	UserFieldName string
+	// Corresponds to the name of the field inside a http(2) request header
+	// to tell the upstream server about the user's groups
+	GroupsFieldName string
+	// The separator string used for concatenating multiple group names in a groups header field's value
+	GroupSeparator string
+}
+
+type AuthnConfig struct {
```
s-urbaniak: Same as above. Also, we should clarify the difference of `AuthnConfig` and `AuthnHeaderConfig` in the godoc.

pkg/authn/delegating.go
```diff
@@ -0,0 +1,28 @@
...
+		Anonymous:    false, // always require authentication
+		CacheTTL:     2 * time.Minute,
+		ClientCAFile: authn.X509.ClientCAFile,
+		TokenAccessReviewClient: client,
+	}
+
+	// if client == nil {
+	// 	return nil, errors.New("no client provided, cannot use webhook authentication")
+	// }
+	// authenticatorConfig.TokenAccessReviewClient = client
```
s-urbaniak: Let's remove commented code.

pkg/authn/oidc.go
```diff
@@ -0,0 +1,40 @@
...
+		GroupsPrefix:         config.GroupsPrefix,
+		SupportedSigningAlgs: config.SupportedSigningAlgs,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	authenticator := bearertoken.New(tokenAuthenticator)
+
+	return authenticator, nil
```
s-urbaniak: nit: reduce the two lines to `return bearertoken.New(tokenAuthenticator), nil`

pkg/authz/auth.go
```diff
@@ -0,0 +1,102 @@
...
+	"time"
+
+	"github.com/brancz/kube-rbac-proxy/pkg/authn"
+
+	"k8s.io/apiserver/pkg/authorization/authorizer"
+	"k8s.io/apiserver/pkg/authorization/authorizerfactory"
+	authorizationclient "k8s.io/client-go/kubernetes/typed/authorization/v1beta1"
+)
+
+type AuthzConfig struct {
```
s-urbaniak: let's add some gocode here.

pkg/authz/auth.go
```diff
@@ -0,0 +1,102 @@
...
+	ResourceAttributes     *ResourceAttributes
+	ResourceAttributesFile string
+}
+
+type ResourceAttributes struct {
```
s-urbaniak: see above re: godoc.

pkg/authz/auth.go
```diff
@@ -0,0 +1,102 @@
...
+	Namespace   string `json:"namespace,omitempty"`
+	APIGroup    string `json:"apiGroup,omitempty"`
+	APIVersion  string `json:"apiVersion,omitempty"`
+	Resource    string `json:"resource,omitempty"`
+	Subresource string `json:"subresource,omitempty"`
+	Name        string `json:"name,omitempty"`
+}
+
+type AuthConfig struct {
```
s-urbaniak: godoc :-)

pkg/proxy/proxy.go
```diff
@@ -0,0 +1,202 @@
...
+	yaml "gopkg.in/yaml.v2"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apiserver/pkg/authentication/authenticator"
+	"k8s.io/apiserver/pkg/authentication/user"
+	"k8s.io/apiserver/pkg/authorization/authorizer"
+	apirequest "k8s.io/apiserver/pkg/endpoints/request"
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+type AuthInterface interface {
```
s-urbaniak: please backport the changes done in https://github.com/brancz/kube-rbac-proxy/pull/14#discussion_r209924958. The interface was removed for, returning plain `(*kubeRBACProxyAuth, error)` instead.

pkg/authn/delegating.go
piotrmsc: Yes, my mistake, forgot to uncomment... I have moved the check to the beginning of func

main.go
```diff
@@ -106,38 +111,66 @@ func main() {
...
+			tokenClient := kubeClient.AuthenticationV1beta1().TokenReviews()
+			authenticator, err = authn.NewDelegatingAuthenticator(tokenClient, cfg.auth.Authentication)
+			if err != nil {
+				glog.Fatalf("Failed to instantiate delegating authenticator: %v", err)
+			}
 		}
 	}
 
-	auth, err := auth.BuildAuthHandler(kubeClient, &cfg.auth)
+	if kubeClient == nil && reflect.ValueOf(kubeClient).IsNil() {
```
s-urbaniak: Is this check really necessary? `initKubeConfig` should err if there is no kubeClient available.

main.go
```diff
@@ -106,38 +111,66 @@ func main() {
...
-			glog.Fatalf("Failed to read resource-attribute file: %v", err)
+			glog.Fatalf("Failed to instantiate OIDC authenticator: %v", err)
 		}
 
-		err = yaml.Unmarshal(b, &cfg.auth.Authorization.ResourceAttributes)
-		if err != nil {
-			glog.Fatalf("Failed to parse resource-attribute file content: %v", err)
+	} else {
+		//Use Delegating authenticator
+		if kubeClient != nil && !reflect.ValueOf(kubeClient).IsNil() {
```
s-urbaniak: see above. I think this check is not necessary.

examples/oidc/deployment.yaml
brancz: let's change this now already to quay.io/brancz/kube-rbac-proxy:v0.4.0

brancz:
This lgtm :+1: thanks a lot for all your hard work!

PR #14, opened 2018-08-14
Extract auth module

brancz:
This pull request extracts the authentication parts with Kubernetes into a re-usable module to be importable by other projects.

@s-urbaniak @squat @mxinden

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package main
+package auth
 
 import (
 	"errors"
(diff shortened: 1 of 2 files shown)


auth/auth.go
```diff
@@ -36,6 +36,14 @@ import (
 	authorizationclient "k8s.io/client-go/kubernetes/typed/authorization/v1beta1"
 )
 
+type AuthInterface interface {
```
s-urbaniak: nit `AuthInterface interface` stutters, can we call this `Authorizer`?

brancz: It encapsulates both authentication and authorization, so `Authorizer` would not give the right notion, I think.

squat: I agree with Frederic. Like anything there can be exceptions to these naming guidelines. One thing is I always find it funny when types have the type in the name. We could call this simply `Auther`.

auth/auth.go
```diff
@@ -76,6 +84,58 @@ type AuthConfig struct {
 	Authorization  *AuthzConfig
 }
 
+func (c *AuthConfig) DeepCopy() *AuthConfig {
+	res := &AuthConfig{
+		Authentication: &AuthnConfig{
+			X509: &X509Config{
+				ClientCAFile: c.Authentication.X509.ClientCAFile,
+			},
+			Header: &AuthnHeaderConfig{
```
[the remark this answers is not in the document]
brancz: You're absolutely right. I think on line 89 we should just not create any deep structures and just build them as we already do later in the code. Nice catch!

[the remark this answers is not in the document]
brancz: Fixed.

auth/auth.go
s-urbaniak: If it encapsulates two separate concerns, it smells like the interface is too big.

Just asking: Do we need to expose this interface at all? Why not simply returning `(*kubeRBACProxyAuth, error)`? The interface declaration could be anonymous, ensuring compile time safety, the caller could then type cast to `Authorizer` and/or `Authenticator` at will.

All of this is type bike shedding though, ignore it at will ;-)

brancz: I like `Auther`, let's go with that.

brancz: @s-urbaniak agreed. I removed it.

PR #17, opened 2018-09-26
*: Add rewrite functionality of SubjectAccessReview requests

brancz:
This PR adds rewrite functionality based on query parameters of an URL. For example, when configured to the `namespace` query parameter then the value passed in a request can be templated in the SubjectAccessReview done against the Kubernetes API. This allows SubjectAccessReviews to be dependent on the parameter passed. For example this can be used with the [prom-label-proxy](https://github.com/openshift/prom-label-proxy) to validate that a requesting user (or Prometheus server) is allowed to access metrics for a certain namespace, and if so it is passed onto the label proxy through, which then enforces the PromQL query to have the respective namespace in the query.

If a rewrite by parameter is configured but no parameter is given, then the request is returned as a `BadRequest`.

@s-urbaniak @metalmatze

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,6 +20,7 @@ import (
 	"crypto/tls"
 	stdflag "flag"
 	"fmt"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"net/http/httputil"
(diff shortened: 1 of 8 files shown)


main.go
```diff
@@ -61,6 +63,10 @@ type tlsConfig struct {
 	cipherSuites []string
 }
 
+type Configfile struct {
```
s-urbaniak: nit: do we need to make this public? For json parsing only this can be left private.

pkg/proxy/proxy.go
```diff
@@ -59,23 +58,31 @@ func (h *kubeRBACProxy) Handle(w http.ResponseWriter, req *http.Request) bool {
...
-		return false
-	}
-	if authorized != authorizer.DecisionAllow {
-		msg := fmt.Sprintf("Forbidden (user=%s, verb=%s, resource=%s, subresource=%s)", u.GetName(), attrs.GetVerb(), attrs.GetResource(), attrs.GetSubresource())
+	allAttrs := h.authorizerAttributesGetter.GetRequestAttributes(u, req)
+	if len(allAttrs) == 0 {
+		msg := fmt.Sprintf("Bad Request. The request or configuration is malformed.")
 		glog.V(2).Info(msg)
-		http.Error(w, msg, http.StatusForbidden)
+		http.Error(w, msg, http.StatusBadRequest)
```
s-urbaniak: :+1:

main.go
brancz: No sorry actually it was not exported at first and I was debugging something. Will change to be unexported.

PR #21, opened 2018-11-16
Remove hardcoded amd64 to allow multi-arch builds

jeffdyoung:
Same as this: https://github.com/openshift/kube-rbac-proxy/pull/4


brancz:
lgtm :+1:

PR #27, opened 2019-01-21
Check errors and log them

metalmatze:
When starting a server we should check their return values and log them.
Additionally, we should think about using something like `run.Group` to terminate all goroutines if one server fails.

/cc @brancz @squat @mxinden


brancz:
Lgtm and +1 and rungroup, vet, staticcheck etc.

PR #30, opened 2019-01-23
Use run.Group to synchronize server goroutines

metalmatze:
This moves the servers to run in synchronized goroutines rather than running them without coordination.

Depends on #29 

/cc @brancz


brancz:
Nice work :+1:

PR #31, opened 2019-01-23
*: Write Changelog and bump version to v0.5.0

metalmatze:
We should do another release to push minimum fixes to everyone.
After that we can take some time and work on more linting etc.

/cc @brancz


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 ## Next release
 
+## 0.5.0 / 2019-01-23
```
brancz: Nothing here is a feature, so let's make this 0.4.1

metalmatze:
Good point. I've forced pushed a changed version

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 ## Next release
 
+## 0.4.1 / 2019-01-23
+
+* [ENHANCEMENT] Use golang.org/x/net http2 server. #29
+* [ENHANCEMENT] Update Kubernetes to 1.13.2 #28
+* [ENHANCEMENT] Make multi-arch builds possible. #21
+* [BUGFIX] Log when server isn't able to start. #27
```
brancz: This also needs.

```
* [BUGFIX] Set user specified TLS configuration when explicit TLS certificates are provided.
```

brancz:
Let's move forward with this

PR #33, opened 2019-02-12
Add support for x-forwarded-for header

josefkarasek:
This header is commonly used by software that can be configured
to a _behind a proxy_ mode. Requests not only have to present
specific http headers, request also have to come from a range
of specific addresses, which makes header spoofing attacks
more difficult.

### DIFF
--- a/main.go
+++ b/main.go
@@ -197,8 +197,15 @@ func main() {
 	}
 
 	upstreamTransport, err := initTransport(cfg.upstreamCAFile)
+
 	if err != nil {
 		glog.Fatalf("Failed to set up upstream TLS connection: %v", err)
+	}
+
+	forwardedFor, err := getOutboundIP(upstreamURL.Host)
+
+	if err != nil {
+		glog.Fatalf("Failed to find outbound IP for upstream %s: %v", upstreamURL.Host, err)
 	}
 
 	proxy := httputil.NewSingleHostReverseProxy(upstreamURL)
@@ -211,12 +218,6 @@ func main() {
 		}
 
 		if cfg.auth.Authentication.Header.Enabled {
-			forwardedFor, err := getOutboundIP(upstreamURL.Host)
-			if err != nil {
-				glog.Errorf("Failed to find outbound IP for upstream %s: %v", upstreamURL.Host, err)
-				http.Error(w, http.StatusText(http.StatusServiceUnavailable), http.StatusServiceUnavailable)
-				return
-			}
 			req.Header.Set(cfg.auth.Authentication.Header.XForwardedFor, forwardedFor)
 		}
 
--- a/main.go
+++ b/main.go
@@ -352,7 +352,23 @@ func initKubeConfig(kcLocation string) *rest.Config {
 	return kubeConfig
 }
 
+// getOutboundIP returns IP address of the interface used to communicate with upstreamHost
 func getOutboundIP(upstreamHost string) (string, error) {
+	// When net.Dial is used with UDP protocol, all internal data structures are populated
+	// with actual information, but no communication is performed yet (no TCP handshake).
+	// This is the most reliable way to get address of the interface, what will be used.
+	//
+	// There are alternative ways how to get IP address of a local interface.
+	// But none of them reliably provide address of the interface which will actually be
+	// used to talk to upstreamHost:
+	//
+	//   hostname, _ := os.Hostname()
+	//   net.LookupHost(hostname)
+	//
+	//   ifaces, _ := net.Interfaces()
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -201,6 +202,15 @@ func main() {
 			return
 		}
 
+		if cfg.auth.Authentication.Header.Enabled {
+			forwardedFor, err := getOutboundIP(upstreamURL.Host)
+			if err != nil {
+				glog.Errorf("Failed to find outbound IP for upstream %s: %v", upstreamURL.Host, err)
+				return
```
s-urbaniak: if no outbound IP could be found, we are returning from the http handler function immediately. We should at least probably return a 503 here?

josefkarasek: Or... Don't return. The proxy call should also fail.

main.go
```diff
@@ -201,6 +202,16 @@ func main() {
 			return
 		}
 
+		if cfg.auth.Authentication.Header.Enabled {
+			forwardedFor, err := getOutboundIP(upstreamURL.Host)
+			if err != nil {
+				glog.Errorf("Failed to find outbound IP for upstream %s: %v", upstreamURL.Host, err)
+				http.Error(w, "Service currenlty unavailable", http.StatusServiceUnavailable)
```
s-urbaniak: please use the constant text instead: https://github.com/golang/go/blob/go1.11.5/src/net/http/status.go#L134

s-urbaniak: there is also a helper lookup method in https://github.com/golang/go/blob/go1.11.5/src/net/http/status.go#L146

s-urbaniak:
@josefkarasek the README conflicts with the PR merged just before. You might need to merge the changes and resolve the conflict.

main.go
```diff
@@ -339,3 +350,13 @@ func initKubeConfig(kcLocation string) *rest.Config {
 
 	return kubeConfig
 }
+
+func getOutboundIP(upstreamHost string) (string, error) {
+	conn, err := net.Dial("udp", upstreamHost)
```
brancz: This is a really expensive operation, as the upstream host is known, can we perform this once at startup instead of on each request? Then we could also exit hard if we fail to get a local address.

main.go
```diff
@@ -209,6 +210,16 @@ func main() {
 			return
 		}
 
+		if cfg.auth.Authentication.Header.Enabled {
```
josefkarasek: @brancz I don't want `auth.Handle()` to have any knowledge about the outbound address, but as a result this there's Header handling in two different places. Any advice, which way is better? Thank you

brancz: I think the way you did it is fine. We could potentially have auth return the auth information instead of silently attaching it to the request, but that would be outside of this PR anyways.

main.go
```diff
@@ -339,3 +351,13 @@ func initKubeConfig(kcLocation string) *rest.Config {
 
 	return kubeConfig
 }
+
+func getOutboundIP(upstreamHost string) (string, error) {
```
brancz: I personally think this is a bit odd, can we add a comment why this is the way we are retrieving the address? I'm not concerned with the expensiveness anymore as we're just doing it once at startup, but reading this in 3 months I'll want to understand why this is the best way of getting the preferred local IP.

josefkarasek:
@brancz should I squash all commits to one?

brancz:
I think having the commits as they are is perfectly fine. This looks good to me now. I'm just thinking that for #34 we should cut a 0.4.2 release, and then a 0.5 release that includes this. What do you think @s-urbaniak @metalmatze ?

s-urbaniak:
@brancz sounds good for both propositions

metalmatze:
@brancz Yup, sounds good.

main.go
```diff
@@ -209,6 +217,10 @@ func main() {
 			return
 		}
 
+		if cfg.auth.Authentication.Header.Enabled {
+			req.Header.Set(cfg.auth.Authentication.Header.XForwardedFor, forwardedFor)
```
slaskawi, line 221: This looks a bit dangerous. By a mistake you can override some existing headers. Is that intentional?

main.go
slaskawi: I totally agree.

slaskawi (changes requested):
The implementation looks ok. Can I ask to resolve outstanding comments and squash the commits as requested before.

ibihim:
@josefkarasek, are you still actively developing on this branch? Otherwise I would close it within a month.

ibihim:
I will close it as of now. If you want to continue, please reopen the PR. Thx.

PR #34, opened 2019-02-12
Add support for custom upstream CAs

josefkarasek:


main.go
```diff
@@ -331,3 +341,36 @@ func initKubeConfig(kcLocation string) *rest.Config {
...
 }
+
+func initTransport(upstreamCAFile string) (http.RoundTripper, error) {
+	if upstreamCAFile == "" {
+		return nil, nil
+	}
+
+	rootPEM, err := ioutil.ReadFile(upstreamCAFile)
+	if err != nil {
+		return nil, err
```
s-urbaniak: let's add some more context here, i.e. `fmt.Errorf("error reading upstream CA file: %v", err)`.

main.go
```diff
@@ -331,3 +341,36 @@ func initKubeConfig(kcLocation string) *rest.Config {
...
+		Proxy: http.ProxyFromEnvironment,
+		DialContext: (&net.Dialer{
+			Timeout:   30 * time.Second,
+			KeepAlive: 30 * time.Second,
+			DualStack: true,
+		}).DialContext,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
```
s-urbaniak: do you mind add a small comment from what Go version these default values are sourced from?

main.go
```diff
@@ -331,3 +341,36 @@ func initKubeConfig(kcLocation string) *rest.Config {
...
+	}
+
+	rootPEM, err := ioutil.ReadFile(upstreamCAFile)
+	if err != nil {
+		return nil, err
+	}
+
+	roots := x509.NewCertPool()
+	if ok := roots.AppendCertsFromPEM([]byte(rootPEM)); !ok {
+		return nil, err
```
s-urbaniak: this block doesn't have an `err` instance at hand, just an `ok == false`, if any certificates were not parsed successfully. I think this block needs an explicit `return nil, errors.New("...")`.

main.go
```diff
@@ -331,3 +341,36 @@ func initKubeConfig(kcLocation string) *rest.Config {
 
 	return kubeConfig
 }
+
+func initTransport(upstreamCAFile string) (http.RoundTripper, error) {
+	if upstreamCAFile == "" {
+		return nil, nil
```
s-urbaniak: if there is no upstream CA file here, we are returning `nil` here, which effectively uses the default transport later on. For brevity, and because this is an init method, let's return the default transport here to make it explicit.

josefkarasek:
> @josefkarasek looking good so far, i have a few comments. Do you also mind to add a small unit test for the init method?

@s-urbaniak Let's move it to a utils package first? Am thinking about unit testing `package main`.
`main.go` and `main_test.go`?

s-urbaniak:
@josefkarasek it is not quite idiomatic to have general `utils` packages in golang. I'd rather, in this case, factor this single method in a single file `transport.go` and create a `transport_test.go` file here too, which implements the unit test. For simplicity, I'd leave this here in the `main` package.

main.go
```diff
@@ -331,3 +341,36 @@ func initKubeConfig(kcLocation string) *rest.Config {
...
+	if err != nil {
+		return nil, err
+	}
+
+	roots := x509.NewCertPool()
+	if ok := roots.AppendCertsFromPEM([]byte(rootPEM)); !ok {
+		return nil, err
+	}
+
+	transport := &http.Transport{
```
squat: Do we care about ALL of these settings or can we use the default transport and simply set the TLSClientConfig field?

josefkarasek: Yes we care, they're set for a reason. You might run into troubles, eg run out of file descriptors etc.

transport_test.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"net/http"
+	"reflect"
+	"testing"
+)
+
+func TestInitTransportWithDefault(t *testing.T) {
+	roundTripper, err := initTransport("")
+	if err != nil {
+		t.Errorf("initTransport(\"\") is legitimate. Should return http.DefaultTransport. But returned %v", err)
```
s-urbaniak: nit, a more idiomatic test failure would be:
```
t.Errorf("want err to be nil, but got %v", err)
```

transport_test.go
```diff
@@ -0,0 +1,28 @@
...
+	}
+	if eq := reflect.DeepEqual(http.DefaultTransport, roundTripper); !eq {
```
s-urbaniak: I wouldn't deep equal with the stdlib one, I'd say
```
if roundTripper == nil {
  t.Errorf("expected roundtripper, got nil")
}
```

transport_test.go
```diff
@@ -0,0 +1,28 @@
...
+		t.Error("initTransport(\"\") did not return http.DefaultTransport")
+	}
+}
+
+func TestInitTransportWithCustomCA(t *testing.T) {
+	roundTripper, err := initTransport("test/ca.pem")
+	if err != nil {
+		t.Errorf("initTransport failed with: %v", err)
```
s-urbaniak: see above, also this should `return` from the test.

transport_test.go
```diff
@@ -0,0 +1,28 @@
...
+	}
+	transport := roundTripper.(*http.Transport)
+	if transport.TLSClientConfig.RootCAs == nil {
+		t.Error("custom CA cert want't applied")
```
s-urbaniak: `t.Error("expected root CA to be set, got nil")`

main.go
josefkarasek: @squat I misunderstood you.

We can use http.DefaultTransport direclty. I just wanted to be explicit, instead of changing types between `http.RoundTripper` and `http.Transport`.

But my understanding is that this would work:
```
	transport := http.DefaultTransport.(*http.Transport)
	transport.TLSClientConfig = &tls.Config{RootCAs: roots}

	return transport, nil
```

transport_test.go
josefkarasek: If I understand the logic correctly, here we should `return` too?

s-urbaniak: yes :+1:

transport.go
```diff
@@ -0,0 +1,47 @@
...
+		return nil, fmt.Errorf("error reading upstream CA file: %v", err)
+	}
+
+	roots := x509.NewCertPool()
+	if ok := roots.AppendCertsFromPEM([]byte(rootPEM)); !ok {
+		return nil, errors.New("error parsing upstream CA certificate")
+	}
+
+	transport := http.DefaultTransport.(*http.Transport)
+	transport.TLSClientConfig = &tls.Config{RootCAs: roots}
```
s-urbaniak: the previous version was better IMO. Here, you are modifying the runtime wide `DefaultTransport` which might have side effects.

josefkarasek: oh crap it's a pointer... you're absolutely right

PR #35, opened 2019-02-14
Fix proxy test

josefkarasek:
apiserver.authenticator in [kube 1.11](https://github.com/kubernetes/kubernetes/blob/release-1.11/staging/src/k8s.io/apiserver/pkg/authentication/authenticator/interfaces.go#L47)
vs
[kube 1.13](https://github.com/kubernetes/kubernetes/blob/release-1.13/staging/src/k8s.io/apiserver/pkg/authentication/authenticator/interfaces.go#L46)

Which is what's now in glide.yaml.

make test output:
```
>> running all tests
# install test dependencies
# run the tests
?   	github.com/brancz/kube-rbac-proxy	[no test files]
?   	github.com/brancz/kube-rbac-proxy/pkg/authn	[no test files]
?   	github.com/brancz/kube-rbac-proxy/pkg/authz	[no test files]
ok  	github.com/brancz/kube-rbac-proxy/pkg/proxy	0.019s
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,10 @@ grpcc-container:
 
 test:
 	@echo ">> running all tests"
+	# install test dependencies
 	@go test -i $(PKGS)
+	# run the tests
+	@go test  $(PKGS)
 
 generate: embedmd
 	@echo ">> generating examples"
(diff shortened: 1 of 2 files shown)


s-urbaniak:
nice :+1: thank you for fixing this!

Issue #40, opened 2019-06-07
Sweet32 still allowed after upgrading to kube-rbac-proxy 0.4.1 and setting minimum TLS version to 1.2


GhostViz:
after upgrading kube-rbac-proxy to version 0.4.1 and setting TLS Minimum Version to 1.2
```
 --tls-min-version=VersionTLS12 
```

`openssl s_client -connect [redacted]:9100 -cipher DES-CBC3-SHA` is still able to connect and verify. 

```
SSL handshake has read 2025 bytes and written 425 bytes
---
New, TLSv1/SSLv3, Cipher is DES-CBC3-SHA
Server public key is 2048 bit
Secure Renegotiation IS supported
Compression: NONE
Expansion: NONE
SSL-Session:
    Protocol  : TLSv1.2
    Cipher    : DES-CBC3-SHA
```

`openssl s_client -connect [redacted]:9100 -cipher ECDHE-RSA-DES-CBC3-SHA` has teh same behavior as listed above, of course with the cipher being ECDHE-RSA-DES-CBC3-SHA instead of DES-CBC3-SHA.

For more context  on how I am using kube-rbac-proxy, it is being used as part of an openshift monitoring deployment in front of node_exporter.

brancz:
Thanks for reporting! Yes we should definitely fix this.

brancz:
For what it's worth, the ciphers used you can configure with the `--tls-cipher-suites` flag, but we should make TLS 1.3 the new default and recompile with latest go version.

brancz:
For clarification, which version of OpenShift are you using and how is the kube-rbac-proxy started? It should be running only safe cipher suites.

GhostViz:
Using Openshift version 3.11 with the Openshift Monitoring GAP stack deployed using the openshift-monitoring playbook from openshift-ansible.  By default it uses the kube-rbac-proxy 0.3.1 image from quay. I pulled the 0.4.1 image and edited only the image tag in the deployment. 0.3.1 to 0.4.1.

brancz:
There is no proper support for TLS 1.3 in go yet, so we will have to wait for that (it can be enabled but only via a debug flag). Regarding the cipher suites, is there anything you can't achieve with the existing `--tls-cipher-suites` flag?

GhostViz:
[...]

PR #41, opened 2019-06-11
main.go: Allow setting min TLS to 1.3

brancz:
Fixes parts of https://github.com/brancz/kube-rbac-proxy/issues/40

@s-urbaniak @metalmatze @paulfantom @squat


brancz:
I've decided not to go ahead with this. Instead we should migrate to https://github.com/kubernetes/component-base and use TLS config flags from there, including selecting TLS min version, cipher suites, etc.

PR #43, opened 2019-06-13
*: Move to go mods

brancz:
While working on a few things go mods were a bit in my way (or rather the lack thereof), so I decided to just go ahead and make the switch.

@s-urbaniak @metalmatze @paulfantom @squat


metalmatze:
You probably want to set `GO111MODULE=on` inside the Makefile for the `go build` command, at least until its enabled by default in Go 1.13.

brancz:
@metalmatze I split it into two commits now, so it's easier to see. I already set the env variable in `.travis.yml`.

PR #45, opened 2019-06-17
 *: watch tls certificate changes

s-urbaniak:
TODOs:
- [ ] test on real cluster

/cc @brancz @metalmatze @paulfantom


s-urbaniak:
setting as WIP, as I am losing fsnotify events after `REMOVE` events. Viper has a reference implementation for this: https://github.com/spf13/viper/pull/415

brancz:
Let me know when it's ready for review. It may very well be that the config watchers are not 100% correct today. Let's not worry about those right now, and just do it properly here :)

s-urbaniak:
@brancz agreed :+1: I scanned the prometheus config reloader and I believe it is defunct right now (at least for fsnotify events). I think viper's method is reasonable and that is what I am porting over.

I am finalizing the tests and verifying the design with @lucab. Stay tuned :-)

s-urbaniak:
@brancz this is ready for an initial pass. We have the following situation:

Simply watching the cert/key paths won't work, because in k8s land an atomic configmap/secret swap is done using a double symlink approach:
```
..dir_rev1
..data -> ..dir1
tls.crt -> ..data/tls.crt
tls.key -> ..data/tls.key
```
k8s atomically swaps `tls.crt`/`tls.key` by creating `dir_rev2` and overwriting the `..data` symlink:
```
..dir_rev2
..data -> ..dir_rev2
tls.crt -> ..data/tls.crt
tls.key -> ..data/tls.key
```
Since golang's fsnotify library always evaluates symlinks as per https://github.com/fsnotify/fsnotify/blob/c2828203cd70a50dcccfb2761f8b1f8ceef9a8e9/kqueue.go#L192 we only get a `REMOVE` event for `..dir_rev1/tls.crt` and `..dir_rev1/tls.key`. As those targets are gone afterwards, we don't receive any follow-up events.

viper's approach linked above does
1. evaluate the symlink
2. watch for changes on the whole directory
3. recognize if the symlink target changed
4. if so, reload the content

This is a sound approach, but tracking the symlink target evaluation implies state tracking which we can avoid by simply watching the intermediate symlink `..data`.

Whenever k8s updates the configmap, a `CREATE` event is triggered on the `..data` symlink, signalling us the necessary event to reload the certificates. We don't need to track state in the watcher.

The only problem is, that we cannot fsnotify on `..data`, as golang's fsnotify library evaluates symlinks ;-) 

As a workaround, we watch for events in the whole directory (as viper does), and filter out events relating to `..data`.

This implementation also includes the possibility to track changes on the certificates directly in case they are simple files and not symlinks for brevity.

pkg/tls/reloader.go
```diff
@@ -0,0 +1,172 @@
...
+			return fmt.Errorf("error adding watchable: %v", err)
+		}
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+
+		case event := <-watcher.Events:
```
s-urbaniak: I wonder if we, in addition should add a ticker anyways. I am not sure we have 100% guarantees to catch all file system events, cc @lucab.

brancz: Curious: for this very case, why not just read the files once per minute and apply if they changed? For other cases I agree that reloading as quickly as possible when available makes perfect sense, but here we have a lot of time to apply cert rotation anyways no?

s-urbaniak: sounds good yes, this will simplify a lot of things and we don't have to think about edge cases :+1: let me implement a ticker-only based approach here and reuse the implemented logic here for prometheus-config-reloader.

s-urbaniak:
this is historic and can be closed

PR #46, opened 2019-06-19
Use go env GOARCH in Makefile

yselkowitz:
On aarch64, GOARCH is 'arm64'.  Instead of making this even longer, just get GOARCH from go itself.


brancz:
Thanks!

PR #47, opened 2019-06-20
*: watch tls certificate changes

s-urbaniak:
TODOs:
- [ ] check on a running cluster

This is an implementation as discussed in https://github.com/brancz/kube-rbac-proxy/pull/45#discussion_r295295712, ready for an initial pass.

I am keeping the other PR open to scavenge the code for the prometheus config reloader implementation.

### DIFF
--- a/pkg/tls/reloader.go
+++ b/pkg/tls/reloader.go
@@ -1,0 +1,122 @@
+/*
+Copyright 2017 Frederic Branczyk All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tls
+
+import (
+	"bytes"
+	"context"
+	"crypto/tls"
+	"fmt"
+	"io/ioutil"
+	"sync"
+	"time"
+
+	"github.com/golang/glog"
+)
+
+// CertReloader is the struct that parses a certificate/key pair,
+// providing a goroutine safe GetCertificate method to retrieve the parsed content.
+//
+// The GetCertificate signature is compatible with https://golang.org/pkg/crypto/tls/#Config.GetCertificate
+// and can be used to hot-reload a certificate/key pair.
+//
+// For hot-reloading the Watch method must be started explicitly.
+type CertReloader struct {
+	certPath, keyPath string
+	interval          time.Duration
+
+	mu              sync.RWMutex // protects the fields below
+	cert            *tls.Certificate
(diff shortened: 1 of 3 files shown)


pkg/tls/reloader.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+}
+
+func (r *CertReloader) reload() error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	var err error
+
+	crt, err := ioutil.ReadFile(r.certPath)
```
brancz: Unlikely but I think we should check both errors here

pkg/tls/reloader.go
```diff
@@ -0,0 +1,125 @@
...
+	key, err := ioutil.ReadFile(r.keyPath)
+
+	if err != nil {
+		return fmt.Errorf("error loading certificate: %v", err)
+	}
+
+	h := fnv.New64a()
+	_, err = h.Write(crt)
```
brancz: Is hashing really necessary? Seems like we should just be able to compare the bytes of the certs themselves.

pkg/tls/reloader.go
s-urbaniak: oh, doh, i wanted to achieve a running err check, silly me, i'll add an err check here too obviously, good catch :+1:

pkg/tls/reloader.go
s-urbaniak: sounds good :+1:

s-urbaniak: small complication, I rechecked, and the parsing function stores DEM bytes as per https://golang.org/src/crypto/tls/tls.go?s=6337:6408#L193, so comparing the raw file content is not applicable.

Does comparing the https://golang.org/pkg/crypto/tls/#Certificate.Certificate slice _after_ parsing sound good to you?

brancz: Can’t we save the raw content bytes in the same fashion as we do the hash right now?

s-urbaniak: we definitely can, i'll add that :+1:

s-urbaniak:
@brancz addressed comments, PTAL

pkg/tls/reloader.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+
+		if err := r.reload(); err != nil {
+			return fmt.Errorf("reloading failed: %v", err)
+		}
+	}
+}
+
+func (r *CertReloader) reload() error {
+	r.mu.Lock()
```
brancz: This lock seems a little excessive, let’s just lock around the actual replacement, once we’ve ensured we really do need to replace and just read lock around reading the “current” cert bytes (as in the if statement doing bytes.Equal)

s-urbaniak: ack, i've been thinking the same, but reasoned that this locks once per minute, on average. agreed on the finer granularity :+1:

brancz: if this was just for exporters I'd be fine with that, but we're guarding Prometheus with this, which gets frequent query requests, so I'd prefer to do this at this point already

s-urbaniak: that's perfectly reasonable, addressed, ptal

brancz:
lgtm :+1:

PR #55, opened 2019-12-03
*: Move from glog to klog

brancz:
I noticed that we don't actually log much even when bumping verbosity, I noticed that this is because we haven't moved from glog to klog. After this PR plenty more log lines appear when bumping to higher log verbosity.

@s-urbaniak @paulfantom @LiliC


s-urbaniak:
lgtm :+1:

PR #59, opened 2020-01-14
Add stale cache option

nabokihms:
### Description
This PR adds the new command-line option - `--stale-cache-interval`. 
When option set, it's enabled additional lru cache to store token review results.

### Motivation
kube-rbac-proxy is a handy tool to protect your metrics. But when apiserver not responding, you are completely lost the possibility to access any metrics. The stale cache makes kube-rbac-proxy works without apiserver for some time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ Usage of _output/kube-rbac-proxy:
       --ignore-paths strings                        Comma-separated list of paths against which kube-rbac-proxy will proxy without performing an authentication or authorization check. Cannot be used with --allow-paths.
       --insecure-listen-address string              The address the kube-rbac-proxy HTTP server should listen on.
       --kubeconfig string                           Path to a kubeconfig file, specifying how to connect to the API server. If unset, in-cluster configuration will be used
+      --stale-cache-interval string	                The interval to keep auth request review result for in case of unconsciousness of apiserver (default 0)
       --log_backtrace_at traceLocation              when logging hits line file:N, emit a stack trace (default :0)
       --log_dir string                              If non-empty, write log files in this directory
       --log_file string                             If non-empty, use this log file
--- a/README.md
+++ b/README.md
@@ -41,7 +41,6 @@ Usage of _output/kube-rbac-proxy:
       --ignore-paths strings                        Comma-separated list of paths against which kube-rbac-proxy will proxy without performing an authentication or authorization check. Cannot be used with --allow-paths.
       --insecure-listen-address string              The address the kube-rbac-proxy HTTP server should listen on.
       --kubeconfig string                           Path to a kubeconfig file, specifying how to connect to the API server. If unset, in-cluster configuration will be used
-      --stale-cache-interval string                 The interval to keep auth request review results for in case of unavailability of kube-apiserver (default 0)
       --log_backtrace_at traceLocation              when logging hits line file:N, emit a stack trace (default :0)
       --log_dir string                              If non-empty, write log files in this directory
       --log_file string                             If non-empty, use this log file
@@ -57,6 +56,7 @@ Usage of _output/kube-rbac-proxy:
       --secure-listen-address string                The address the kube-rbac-proxy HTTPs server should listen on.
       --skip_headers                                If true, avoid header prefixes in the log messages
       --skip_log_headers                            If true, avoid headers when opening log files
+      --stale-cache-interval duration               The interval to keep auth request review results for in case of unavailability of kube-apiserver.
       --stderrthreshold severity                    logs at or above this threshold go to stderr (default 2)
       --tls-cert-file string                        File containing the default x509 Certificate for HTTPS. (CA cert, if any, concatenated after server cert)
       --tls-cipher-suites strings                   Comma-separated list of cipher suites for the server. Values are from tls package constants (https://golang.org/pkg/crypto/tls/#pkg-constants). If omitted, the default Go cipher suites will be used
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -37,6 +37,7 @@ Usage of _output/linux/amd64/kube-rbac-proxy:
       --config-file string                          Configuration file to configure kube-rbac-proxy.
       --insecure-listen-address string              The address the kube-rbac-proxy HTTP server should listen on.
       --kubeconfig string                           Path to a kubeconfig file, specifying how to connect to the API server. If unset, in-cluster configuration will be used
+      --stale-cache-interval string	                The interval to keep auth request review result for in case of unconsciousness of apiserver (default 0)
```
brancz: Wouldn't it make sense, in order to tolerate some downtime, to have this cache default to let's say 10 minutes? Also the name doesn't reflect to me what this means. I would rather use the same naming as the code uses, something like `--token-cache-ttl`. What do you think?

pkg/proxy/proxy.go
```diff
@@ -47,29 +49,55 @@ type kubeRBACProxy struct {
...
 // If the authn fails, a 401 error is returned. If the authz fails, a 403 error is returned
 func (h *kubeRBACProxy) Handle(w http.ResponseWriter, req *http.Request) bool {
+	token := getTokenFromRequest(req)
+	if token == "" {
+		klog.Errorf("Unable to get token from request")
+		http.Error(w, "Unauthorized", http.StatusUnauthorized)
+		return false
+	}
 	// Authenticate
 	u, ok, err := h.AuthenticateRequest(req)
```
brancz: Wouldn't we instead want to attempt to read the previous response from the cache and only if stale or not existant move forward with the real request? The current strategy looks like we're always going to attempt the real request. Am I missing something?

nabokihms: In my thoughts, we have to use this cache only in case of kube-apiserver unconscious.

For example:
If you haven't high available control-plane (or, as people are calling it, multi-master cluster), you will have holes in graphs every time kube-apiserver will go down. 

Prometheus requests metrics periodically. If Prometheus was allowed to scrape targets when kube-apiserver was healthy, we would like to allow him to scrape targets for some period (like 15min or 30min) even if kube-apiserver is not responding.

However, if kube-apiserver responds, we would like to get the most up to date information about the identity and access rights of the user.

Please correct me if I am wrong.

README.md
nabokihms: I described my thoughts about the purpose of this cache below. For now, I have no idea which name to give to the option.

nabokihms:
I'm awfully sorry. It took me time to ensure everything is working as expected. Now I'm ready to work on this PR. If you are still interested in this feature, let's continue to discuss it.

nabokihms:
I have made the necessary changes. It seems to work per my smoke tests. 

Give me some time to test it in real clusters, and then we will be able to merge this.

nabokihms:
@s-urbaniak I tested this patch in production. Everything should be fine now. Is there something left for me to work on?

pkg/proxy/proxy.go
ibihim: Hm, yes. I had the initial assumption as @brancz had, when look at your PR title and I was super excited :smile: 

After the comment, I understand your use case, but you mention this clearly in the README.md.

pkg/proxy/proxy.go
```diff
@@ -67,44 +86,71 @@ func (h *kubeRBACProxy) Handle(w http.ResponseWriter, req *http.Request) bool {
...
-			klog.Errorf("%s: %s", msg, err)
-			http.Error(w, msg, http.StatusInternalServerError)
-			return false
+			_, ok := h.StaleCache.Get(userIdentity)
+			if !ok {
+				msg := fmt.Sprintf("Authorization error (user=%s, verb=%s, resource=%s, subresource=%s)", u.User.GetName(), attrs.GetVerb(), attrs.GetResource(), attrs.GetSubresource())
+				klog.Errorf("%s: %s", msg, err)
+				http.Error(w, msg, http.StatusInternalServerError)
+
+				h.StaleCache.Remove(userIdentity)
```
ibihim, line 133: Why do we try to remove a `userIdentity` that is not in the cache?

ibihim: Makes me wonder if we could rewrite `func (h *kubeRBACProxy) Handle(w http.ResponseWriter, req *http.Request) bool` into something like `func (h *kubeRBACProxy) Handle(w http.ResponseWriter, req *http.Request) (bool, error)` and wrap the cache as a middleware like

```go
func (h *kubeRBACProxy) Handle(w http.ResponseWriter, req *http.Request) (bool) {
    ok, err :=h.handle(w, req)
    if err != nil {
        // Read from cache and potentially return true.

        return false
    }
    if !ok {
        // Remove from cache

        return false
    }

    // Add to cache.

    return true.
}
```

WDYT?

ibihim (commented):
Thank you for your contribution :smile: 

I know that @s-urbaniak is super proud of the e2e test coverage. Do you think it would be possible to add some of those too?

pkg/proxy/proxy.go
nabokihms: Good point, no objections. I will see what I can do with this.

nabokihms:
@ibihim thanks for the suggestions. 

AFAIR, when this PR was opened, there were no e2e tests. I think it should not be a problem to add one.

ibihim:
@nabokihms, hehe. The PR is that old? Thank you, for your patience.

ibihim:
@nabokihms, this PR is getting stale. Do you want to continue to work on it? I would close it, but leave it open to you to reopen it.

ibihim:
I will close it as of now.If you continue to work on it, please reopen it and rebase.

PR #66, opened 2020-04-24
Enable providing SANs for self-signed certs

avorima:


brancz:
Can you explain why you need this? Self signed certs should not be used if you want any type of security. If you actually care then rather issue real certs.

avorima:
Sorry, this PR was a bit short-sighted. You're right, I'll use real certs.

PR #83, opened 2020-08-11
Add '--allow-paths-regex' CLI flag

simonpasquier:


brancz:
What's the concrete use case we have here? I'm just wondering if regex is really the right thing for this or if we should rather use a list of slightly less flexible template strings. I'm just worried that this will be used accidentally incorrectly and then causes potential security problems (which my understanding is is exactly what we're trying to avoid).

simonpasquier:
The concrete use case is when you have several kube-rbac-proxies in front of the same backend with different permissions for each proxy.
For [instance](https://bugzilla.redhat.com/show_bug.cgi?id=1854479), a Thanos querier with 2 kube-rbac-proxy instances:
* kube-rbac-proxy#1 for the `/api/v1/query` and `/api/v1/query_range` endpoints if the user has permissions  to get pod's metrics.
* kube-rbac-proxy#2 for the `/api/v1/rules` endpoint if the user has permissions to get the prometheus rules.

I think it's also fine if we have to specify all paths that are managed by the proxy.

brancz:
I'm all for enabling the use case but I feel getting regexes right in order for a security feature to work doesn't seem like a good idea, I'd be more comfortable to have explicit paths each listed without regex support. That should still fulfill the requirement no?

simonpasquier:
> That should still fulfill the requirement no?
Completely. I'll update the PR accordingly :)

test/e2e/basics.go
```diff
@@ -190,6 +191,90 @@ func testTokenAudience(s *kubetest.Suite) kubetest.TestSuite {
 	}
 }
 
+func testAllowPathsRegexp(s *kubetest.Suite) kubetest.TestSuite {
```
s-urbaniak, line 194: @simonpasquier maybe as a small follow-up: `s/testAllowPathsRegexp/testAllowPaths` ?

PR #91, opened 2020-09-18
Add support for ignore-paths

### LINKED ISSUE #25: Support workloads with http probes
To secure application using kube-rbac-proxy application needs to bind on localhost. In cases when pod is not running in host network, application is not reachable for kubelet to perform healthchecks. 
To my knowledge there is no way to configure kubelet to use service account tokens when performing http probes, making it impossible to use kube-rbac-proxy with healthchecks without adding yet another proxy just for them. 

I see two possible ways to solve it:
* authorize kubelet http probe to "/healthz" (not sure if it's possible, could be risky to allow access nonResourceURLs on all pods)
* Allow to skip authorization for whitelisted endpoints in kube-rbac-proxy

Samze:
Do not perform any auth check if the request path matches any specified by the new ignore-paths flag, just proxy. Closes #25 

I *think* this can live alongside `--allow-paths` but interested to hear your thoughts.

### DIFF
--- a/main.go
+++ b/main.go
@@ -61,6 +61,7 @@ type config struct {
 	tls                   tlsConfig
 	kubeconfigLocation    string
 	allowPaths            []string
+	ignorePaths           []string
 }
 
 type tlsConfig struct {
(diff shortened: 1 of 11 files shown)


main.go
```diff
@@ -228,9 +230,20 @@ func main() {
...
+				ignorePathFound = true
+				break
+			}
+		}
+
+		if !ignorePathFound {
+			ok := auth.Handle(w, req)
+			if !ok {
+				return
+			}
```
s-urbaniak, line 246: When the user configures both `--ignore-paths` and `--allow-paths` it can get quite confusing:

case 1: if `--ignore-paths` is a subset of `--allow-paths`, simply some or all of the allowed paths will be ignored.
case 2: if `--ignore-paths` is not a subset of `--allow-paths` then the former setting has no effect.

Instead, to have clarity for the user, wouldn't it make more sense to have those settings mutually exclusive?

Samze: Yeah I agree complex behaviour and a potential cause of confusion. I've updated the PR to check that only one of these flags can be used.

simonpasquier (commented):
lgtm

s-urbaniak:
This looks great, thank you for the contribution! LGTM too :tada:

PR #98, opened 2020-10-29
cannot find Service Account in pod to build in-cluster rest config

dahaian:
I got an error is "cannot find Service Account in pod to build in-cluster rest config",but I do not know why.


simonpasquier (commented):
Thanks! Small nits otherwise LGTM.

PR #107, opened 2020-12-21
.github/workflows: fix image publication

dgrisonnet:
There is another problem with the publish job: https://github.com/brancz/kube-rbac-proxy/runs/1589263350

While splitting the create-cluster job into the e2e-test and publish jobs, I had forgotten to duplicate the container build.


brancz:
Thank you!

PR #112, opened 2021-03-15
build.yml: wait until images are available

s-urbaniak:
As suggested in https://github.com/brancz/kube-rbac-proxy/issues/108#issuecomment-769755271

cc @paulfantom


s-urbaniak:
Fixes #108

s-urbaniak:
e2e failed with:
```
no kube-rbac-proxy image
the input device is not a TTY
```
i guess we neither need an interactive shell not a tty driver, so i'll try to remove that option.

s-urbaniak:
ok, that seems to do the trick, let's see.

PR #113, opened 2021-03-16
feat: support client certificate authentication

### LINKED ISSUE #82: Request client certificates
Hello. First of all, thank you for this proxy. It works like a charm. 

I have a problem with that there are no client certificates in the request. The problems seem to be related to server tls configuration. I think we need to change tls config option `ClientAuth` to `RequestClientCert` (default value is `NoClientCert`).
```
srv.TLSConfig.ClientAuth = tls.RequestClientCert
```
The main idea behind this is to extract every client cert and validate it using [x509 authenticator](https://github.com/kubernetes/kubernetes/blob/4d182cec03b880b77fb4879793302dfb3663a281/staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go#L86) on the application side.

[Related PR](https://github.com/kubernetes/kubernetes/pull/6190).
Please forgive me if I get it wrong.

nabokihms:
Fixes #82


nabokihms:
@s-urbaniak, some tests failed, but I believe that it is not connected to my PR. The problem persists in the master branch.

test/e2e/clientcertificates/deployment.yaml
```diff
@@ -0,0 +1,40 @@
...
+        app: kube-rbac-proxy
+    spec:
+      serviceAccountName: kube-rbac-proxy
+      containers:
+        - name: kube-rbac-proxy
+          image: quay.io/brancz/kube-rbac-proxy:local
+          args:
+            - "--secure-listen-address=0.0.0.0:8443"
+            - "--upstream=http://127.0.0.1:8081/"
+            - "--auth-token-audiences=kube-rbac-proxy"
```
s-urbaniak, line 23: nit: this test doesn't assert token audiences so i think we can remove this line.

test/e2e/clientcertificates/deployment-wrongca.yaml
```diff
@@ -0,0 +1,40 @@
...
+        app: kube-rbac-proxy
+    spec:
+      serviceAccountName: kube-rbac-proxy
+      containers:
+        - name: kube-rbac-proxy
+          image: quay.io/brancz/kube-rbac-proxy:local
+          args:
+            - "--secure-listen-address=0.0.0.0:8443"
+            - "--upstream=http://127.0.0.1:8081/"
+            - "--auth-token-audiences=kube-rbac-proxy"
```
s-urbaniak, line 23: see the other comment regarding token audiences

s-urbaniak:
@nabokihms thank you for the contribution, this looks great! i have just one nit really, but do you mind to rebase on master? We might have solved the e2e test failures.

test/e2e/clientcertificates/deployment.yaml
nabokihms: Sure, I will remove it.

s-urbaniak:
we still see those flaky e2e and they seem to be unrelated, but this LGTM, thank you!

PR #116, opened 2021-03-22
add static authorization option

### LINKED ISSUE #110: enhancement: static authorization
To reduce api server load I suggest to implement a static authorization scheme. If configured, then:

1. token access reviews would still be performed towards API server
2. subject access reviews would not be issued but rather validated locally against a static configuration

According to @deads2k this helps tremendously in reducing API server load when kube-rbac-proxy is used in high-volume scraping targets like nodes (via node-exporter) or kubelets.

A prototype of a static authorizer is available for initial introspection of the idea at https://github.com/openshift/kube-rbac-proxy/pull/43/files.

The suggestion is to introduce a `static` array into the current configuration struct, i.e.:

```
type Config struct {
	Rewrites               *SubjectAccessReviewRewrites `json:"rewrites,omitempty"`
	ResourceAttributes     *ResourceAttributes          `json:"resourceAttributes,omitempty"`
	ResourceAttributesFile string                       `json:"-"`
	Static                 []StaticAuthorizationConfig   `json:"static,omitempty"`
}

type StaticAuthorizationConfig struct {
	User            UserConfig `json:"user,omitempty"`
	Verb            string     `json:"verb,omitempty"`
	Namespace       string     `json:"namespace,omitempty"`
	APIGroup        string     `json:"apiGroup,omitempty"`
	APIVersion      string     `json:"apiVersion,omitempty"`
	Resource        string     `json:"resource,omitempty"`
	Subresource     string     `json:"subresource,omitempty"`
	Name            string     `json:"name,omitempty"`
	ResourceRequest bool       `json:"resourceRequest,omitempty"`
	Path            string     `json:"path,omitempty"`
}

type UserConfig struct {
	Name   string   `json:"name,omitempty"`
	UID    string   `json:"uid,omitempty"`
	Groups []string `json:"groups,omitempty"`
}
```

This would then allow declaratively to enable static authorization for non-resource requests like so:
```
  config.yaml: |-
    "authorization":
      "static": [
        {
          "user":
            "name": "system:serviceaccount:monitoring:prometheus-k8s"
          "path": "/metrics"
          "verb": "get"
        }
      ]
```

The static authorizer could also be useful to enforce a rewrite. I.e. to authorize a static user and enforce authorization to a specific namespace one could configure:
```
  config.yaml: |-
    "authorization":
      "resourceAttributes":
        "namespace": "{{ .Value }}"
      "rewrites":
        "byQueryParameter":
          "name": "namespace"
      "static": [
        {
          "user":
            "name": "system:serviceaccount:monitoring:prometheus-k8s"
          "namespace": "my-namespace"
        }
      ]
```

I am happy to file a PR but was wondering if the maintainers have overall objections of adding such a functionality here.

cc @simonpasquier @brancz @paulfantom @lilic @deads2k

jan--f:
This add a staticAuthorizer as discussed in #110.

The config of the static authorizer is currently implemented so that any unspecified fields are interpreted as pass any. E.g if a static config is specified like so:
```
config.yaml: |-
    "authorization":
      "static": [
        {
          "path": "/metrics"
         }
      ]
```
any request to `/metrics` is allowed by the static authorizer. Likewise
```
config.yaml: |-
    "authorization":
      "static": [
        {
          "path": "/metrics"
          "verb": "get"
         }
      ]
```
is a bit more strict as it only allows `GET` on `/metrics`. I think this approach is the most flexible and given the typical scope of the rbac proxy should work fine. There is however a distinct potential for users shooting themselves in the foot.

**TODO**

- [ ] documentation

Fixes #110

### DIFF
--- a/examples/static-auth/README.md
+++ b/examples/static-auth/README.md
@@ -1,0 +1,186 @@
+# Static Authorization example
+
+> Note to try this out with minikube, make sure you enable RBAC correctly. Since minikube v0.26.0 the default bootstrapper is kubeadm - which should enable RBAC by default. For older version follow the instructions [here](../minikube-rbac).
+
+RBAC differentiates in two types, that need to be authorized, resources and non-resources. A resource request authorization, could for example be, that a requesting entity needs to be authorized to perform the `get` action on a particular Kubernetes Deployment.
+
+In this example we deploy the [prometheus-example-app](https://github.com/brancz/prometheus-example-app) and want to preotect it with kube-rbac-proxy, just as detailed in the [rewrite example](../rewrite/README.md). In this example however we will avoid the recurring SubjectAccessReview requests to the api server by allowing kube-rbac-proxy to authorize these requests statically. This is configured in the file passed to the kube-rbac-proxy with the `--config-file` flag. Additionally the `--upstream` flag has to be set to configure the application that should be proxied to on successful authentication as well as authorization.
+
+The kube-rbac-proxy itself also requires RBAC access, in order to perform TokenReviews as well as SubjectAccessReviews for requests that are not statically athorized. These are the APIs available from the Kubernetes API to authenticate and then validate the authorization of an entity.
+
+```bash
+$ kubectl create -f deployment.yaml
+```
+
+The content of this manifest is:
+
+[embedmd]:# (./deployment.yaml)
+```yaml
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: kube-rbac-proxy
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: kube-rbac-proxy
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: kube-rbac-proxy
+subjects:
+- kind: ServiceAccount
+  name: kube-rbac-proxy
+  namespace: default
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: kube-rbac-proxy
+rules:
+- apiGroups: ["authentication.k8s.io"]
+  resources:
+  - tokenreviews
+  verbs: ["create"]
+- apiGroups: ["authorization.k8s.io"]
+  resources:
+  - subjectaccessreviews
+  verbs: ["create"]
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: kube-rbac-proxy
+  name: kube-rbac-proxy
+spec:
+  ports:
+  - name: https
+    port: 8443
+    targetPort: https
+  selector:
+    app: kube-rbac-proxy
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kube-rbac-proxy
+data:
+  config-file.yaml: |+
+    authorization:
+      rewrites:
+        byQueryParameter:
+          name: "namespace"
+      resourceAttributes:
+        apiVersion: v1
+        resource: namespace
+        subresource: metrics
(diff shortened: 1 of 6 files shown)


s-urbaniak:
@jan--f let's add a WIP prefix until documentation is done.

pkg/authz/auth.go
```diff
@@ -61,8 +63,27 @@ type ResourceAttributes struct {
...
+// authorization.
+type StaticAuthorizationConfig struct {
+	User            UserConfig
+	Verb            string `json:"verb,omitempty"`
+	Namespace       string `json:"namespace,omitempty"`
+	APIGroup        string `json:"apiGroup,omitempty"`
+	Resource        string `json:"resource,omitempty"`
+	Subresource     string `json:"subresource,omitempty"`
+	Name            string `json:"name,omitempty"`
+	ResourceRequest *bool  `json:"resourceRequest,omitempty"`
```
s-urbaniak, line 76: any reason this is a pointer? i think tristate logic is not really necessary here, either it is a resource request or not.

examples/static-auth/README.md
```diff
@@ -0,0 +1,186 @@
...
+              - group1
+              - group2
+          verb: get
+          namespace: default
+          apiGroup: apps
+          resourceRequest: true
+          resource: namespace
+          subresource: metrics
+          path: /metrics
+```
```
s-urbaniak, line 185: while this example lists all static authorizer fields, in reality this will never occur. in-flight authorizer Attributes are mutually-excluse either resource request or not, see https://github.com/brancz/kube-rbac-proxy/blob/ea07f61c78044445dc7428e4c1f46360e50fc2e8/pkg/proxy/proxy.go#L151-L181.

Hence, for documentation purposes I suggest to document two examples:

```
  config-file.yaml: |+
    authorization:
      static:
        - user:
            name: UserName
            groups:
              - group1
              - group2
          verb: get
          resourceRequest: false
          path: /metrics
```

and

```
  config-file.yaml: |+
    authorization:
      static:
        - user:
            name: UserName
            groups:
              - group1
              - group2
          verb: get
          namespace: default
          apiGroup: apps
          resourceRequest: false
          resource: namespace
```

pkg/authz/auth.go
jan--f: > either it is a resource request or not.

Yes, the intention is to have the third state as "don't care". So user can configure one static auth that allows both resource and non-resource requests.
Not sure whether that is a valid scenario, say "allow all GET requests for a user" but figured going with tri-state wouldn't hurt anything.

examples/static-auth/README.md
jan--f: Yes sounds good. I did consider the mutually exclusivity but forgot to update the docs.

pkg/authz/auth.go
s-urbaniak: > allow all GET requests for a user

this is a non-resource authorization config (aka `resourceRequest = false`), hence my suggestion not to have tristate, which might be confusing :thinking:

examples/static-auth/README.md
s-urbaniak: i will add this in a follow-up, thanks!

pkg/authz/auth.go
jan--f: You are certainly the expert, my thinking was that a GET could potentially be a resource or non-resource. Like in your example here https://github.com/brancz/kube-rbac-proxy/pull/116#discussion_r622903161

s-urbaniak: yeah, in case of GET we always are in "non-resource" mode (aka `resourceRequest = false`), so I think we can safely factor the tristate logic to a pure `true/false` one :+1:

s-urbaniak:
let's close this out in favor of #125

PR #122, opened 2021-04-27
scripts/publish: fetch tags

s-urbaniak:
cc @paulfantom 


paulfantom (approved):
lgtm

I wonder if it wouldn't be simpler to just use the content of `VERSION` file. Let;s park this thought for later though.

s-urbaniak:
> I wonder if it wouldn't be simpler to just use the content of VERSION file. Let;s park this thought for later though.

agreed.

PR #126, opened 2021-05-07
Revert "Makefile: correctly annotate container image designated platform"

s-urbaniak:
Reverts brancz/kube-rbac-proxy#124 as it broke image builds https://github.com/brancz/kube-rbac-proxy/runs/2525412745?check_suite_focus=true

cc @paulfantom


paulfantom (approved):
lgtm

PR #127, opened 2021-05-26
pkg/authn/delegating: dynamically reload client CA

s-urbaniak:
cc @stlaz @deads2k

### DIFF
--- a/pkg/authn/delegating.go
+++ b/pkg/authn/delegating.go
@@ -18,8 +18,7 @@ package authn
 
 import (
 	"errors"
-	"fmt"
-	"io/ioutil"
+	"net/http"
 	"time"
 
 	"k8s.io/apiserver/pkg/authentication/authenticator"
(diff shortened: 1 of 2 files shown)


s-urbaniak:
cc @brancz @paulfantom ptal if you have any free cycles as well :pray:

paulfantom (approved):
nice! lgtm :+1:

pkg/authn/delegating.go
```diff
@@ -62,6 +55,6 @@ func NewDelegatingAuthenticator(client authenticationclient.TokenReviewInterface
 		APIAudiences:                       authenticator.Audiences(authn.Token.Audiences),
 	}
 
-	authenticator, _, err := authenticatorConfig.New()
-	return authenticator, err
+	auth, _, err := authenticatorConfig.New()
```
deads2k, line 58: eh,  if I'm reviewing, I prefer the original.  We don't charge extra for letters and authenticator and authorizer both share auth.

pkg/authn/delegating.go
```diff
@@ -29,29 +28,23 @@ import (
...
 		if err != nil {
 			return nil, err
 		}
 
-		p, err = dynamiccertificates.NewStaticCAContent(authn.X509.ClientCAFile, caBundle)
-		if err != nil {
-			return nil, err
-		}
+		// worker count is ignored
+		go p.Run(1, ctx.Done())
```
deads2k, line 47: I recommend against having New methods start gofuncs.  Instead, you can return an interface that includes both authenticator.Request and a Run and start this in the caller.

You can call `RunOnce` in this method to ensure you have a valid authenticator to begin.

pkg/authn/delegating.go
s-urbaniak: this shadows the "authenticator" package import but i am agnostic as well.

pkg/authn/delegating.go
s-urbaniak: agreed, and i was having a similar thought, but didn't want to complicate things. agreed, let's return an enhanced interface and start outside the constructor :+1:

s-urbaniak: > You can call RunOnce in this method to ensure you have a valid authenticator to begin.
I think we're fine here as the bundle is loaded in the constructor already https://github.com/kubernetes/apiserver/blob/e78fb1931d1a7682240456bf1883759720b97bb1/pkg/server/dynamiccertificates/dynamic_cafile_content.go#L85-L87

Issue #130, opened 2021-07-12
Support regex-based paths ignoring for RBAC authz/authn


knopt:
Hi there!

I'd like to propose a new feature, which would allow to ignore authn/authz for specific paths, based on a regex. This would allow to handle more complex authn/authz logic than `ignore-paths` flag does.

I would suggest to introduce `ignore-regex` flag, which would accept single go-style regex, and exclude any path that matches the regex from authn/authz. This exclusion would be additional to `ignore-paths` exclusion, meaning that if a path matches one of paths from `ignore-paths` it will be excluded as well. Currently a path has to equal to one of paths from `ignore-paths` to be excluded.

Example usage would be specifing `--ignore-regex='/path/(.)*'` to exclude all subpaths from authn/authz.

Please let me know your thoughts. Once we agree on the solution I'll be happy to submit a PR.

Thanks,
Tomasz

brancz:
cc @s-urbaniak

knopt:
@brancz  @s-urbaniak  Do you have any suggestions?

s-urbaniak:
@knopt thank you for the feature suggestion! A couple of comments:

- Instead af allowing full-blown regular expressions in security sensitive proxies I would rather suggest to with a simpler wildcard based approach, i.e. using https://pkg.go.dev/path#Match.
- Introducing interleaving command line parameters can lead to unexpected results. We had a similar discussion in https://github.com/brancz/kube-rbac-proxy/pull/91#discussion_r491900119. Hence, I would like to leave the clear orthogonality between `--ignore-paths` and `--allow-paths`.
- A single expression setting sounds unaligned with the existing `--*-paths` settings

How about simply extending the current `--ignore-paths` and `--allow-paths` cmdline parameters to allow wildcard matches using https://pkg.go.dev/path#Match? This would be backwards compatible and would suffice your concrete use case.

knopt:
@s-urbaniak  thanks for the reply!

1. Using https://pkg.go.dev/path#Match sounds great!
2. Reusing `--ignore-paths` and `--allow-paths`: that makes sense when using Match as all special characters of Match are restricted characters in URI. I guess here we would rely on the fact that set of allowed URI characters (see https://www.ietf.org/rfc/rfc3986.txt) is exclusive to Match special characters. 

So I guess we agree on the solution, I'll try to post the PR in a couple of days

s-urbaniak:
@knopt feel free to try #135

PR #135, opened 2021-07-26
allow path patterns in --allow-paths and --ignore-paths

raptorsun:
This pull request make it possible to use URL patterns matching.
For example, we can use --allow-paths="/pre/fix1/*" to allow URL such as "/pre/fix1/AAA", "/pre/fix1/BBB"

We intent to make it possible to proxy requests using URL suffixes delimited by slashes as a method to pass arguments, such as [this uses case](https://issues.redhat.com/browse/MON-1695) to pass through query of label values to Prometheus. The URL can be /api/v1/label/<name>/values. Because keeping an exhaust list of labels' names is both time-costing and complicated especially when new labels are created during runtime. So we believe allowing pattern match for URL is a good solution.

### DIFF
--- a/main.go
+++ b/main.go
@@ -27,6 +27,7 @@ import (
 	"net/url"
 	"os"
 	"os/signal"
+	"path"
 	"strings"
 	"syscall"
 	"time"
@@ -217,14 +218,31 @@ func main() {
 		klog.Fatal("Cannot use --allow-paths and --ignore-paths together.")
 	}
 
+	for _, pathAllowed := range cfg.allowPaths {
+		_, err := path.Match(pathAllowed, "")
+		if err != nil {
+			klog.Fatalf("Failed to verify allow path: %s", pathAllowed)
+		}
+	}
+
+	for _, pathIgnored := range cfg.ignorePaths {
+		_, err := path.Match(pathIgnored, "")
+		if err != nil {
+			klog.Fatalf("Failed to verify ignored path: %s", pathIgnored)
+		}
+	}
+
 	proxy := httputil.NewSingleHostReverseProxy(upstreamURL)
 	proxy.Transport = upstreamTransport
 	mux := http.NewServeMux()
 	mux.Handle("/", http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		found := len(cfg.allowPaths) == 0
-		for _, path := range cfg.allowPaths {
-			if req.URL.Path == path {
-				found = true
+		for _, pathAllowed := range cfg.allowPaths {
+			found, err = path.Match(pathAllowed, req.URL.Path)
+			if err != nil {
+				return
+			}
+			if found {
 				break
 			}
 		}
@@ -234,9 +252,12 @@ func main() {
 		}
 
 		ignorePathFound := false
-		for _, path := range cfg.ignorePaths {
-			if req.URL.Path == path {
-				ignorePathFound = true
+		for _, pathIgnored := range cfg.ignorePaths {
+			ignorePathFound, err = path.Match(pathIgnored, req.URL.Path)
+			if err != nil {
+				return
+			}
+			if ignorePathFound {
 				break
 			}
 		}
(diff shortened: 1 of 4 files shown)


simonpasquier:
For context, we had a discussion on whether to allow regex matching when implementing the `--allow-paths` flag in #83. At that time it was dismissed since we had no concrete use case.

fpetkovski:
Should we add some sort of a test for this? It was a bit hard to validate the correctness of the change

raptorsun:
I wonder what will be the best way to test a branch the main function. 🤔

Le mar. 27 juil. 2021 à 08:31, Filip Petkovski ***@***.***> a
écrit :

fpetkovski:
You can extract [this](https://github.com/brancz/kube-rbac-proxy/pull/135/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R225-R233) section into a function that returns found/not found and test it in isolation

simonpasquier:
cc @s-urbaniak since it seems to overlap with the discussion happening in #130

s-urbaniak:
oh sorry, i missed this thread (still iterating over concrete mentions after PTO). Are there any concerns using https://pkg.go.dev/path#Match as outlined in https://github.com/brancz/kube-rbac-proxy/issues/130#issuecomment-887348407?

raptorsun:
Updated PR with path pattern match for both --alow-paths and --ignore-paths

main.go
```diff
@@ -27,6 +27,7 @@ import (
 	"net/url"
 	"os"
 	"os/signal"
+	pathmodule "path"
```
s-urbaniak, line 30: any reason we cannot leave this as `"path"`?

main.go
```diff
@@ -223,8 +224,12 @@ func main() {
 	mux.Handle("/", http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		found := len(cfg.allowPaths) == 0
 		for _, path := range cfg.allowPaths {
-			if req.URL.Path == path {
-				found = true
+			found, err = pathmodule.Match(path, req.URL.Path)
+			if err != nil {
+				klog.Fatalf("Allowed path pattern is not correct: %s", path)
```
s-urbaniak, line 229: This will exit the whole process with error code 255 at runtime. Note this is a handler not bootstrapping code. Please simply `return` in this case.

According to https://pkg.go.dev/path#Match:
> The only possible returned error is ErrBadPattern, when pattern is malformed.

We could leverage this and iterate before the handler in the bootstrapping main goroutine:
```
for _, path := range cfg.allowPaths {
  _, err := pathmodule.Match(path, req.URL.Path)
  if err != nil {
    klog.Fatalf("Failed to verify allow path: %s", path)
  }
}
```

main.go
```diff
@@ -235,8 +240,12 @@ func main() {
...
 		ignorePathFound := false
 		for _, path := range cfg.ignorePaths {
-			if req.URL.Path == path {
-				ignorePathFound = true
+			ignorePathFound, err = pathmodule.Match(path, req.URL.Path)
+			if err != nil {
+				klog.Fatalf("Ignored path pattern is not correct: %s", path)
+				return
+			}
+			if ignorePathFound {
```
s-urbaniak, line 248: see comment above

main.go
raptorsun: because "path" is already used as a symbol for variable in the scope, I rename the module import to avoid confusing the compiler.

main.go
raptorsun: thanks, I will replace the klog.Fatalf by a simple return.

main.go
s-urbaniak: I think it is reasonable to rename the local variable to `p`. The usage distance of that variable is bound to the for loop.

raptorsun: ok, i will rename local variable 👍

main.go
raptorsun: check for allowPaths and ignorePaths are added,too.

s-urbaniak:
@raptorsun this change lgtm in general, but please add some test, and at least one e2e test.

raptorsun:
OK, more e2e test will be added.

raptorsun:
E2E tests have been added. 🧪

s-urbaniak:
lgtm, thank you for the contribution :tada:

PR #155, opened 2022-01-06
Add Makefile target for go dep updates

ibihim:
## What

Add `make` target for updating go dependencies.

## Why

- To make it easy to keep the dependencies up to date.
- To make it part of the release process.


s-urbaniak:
let's leverage the make target, and update the godeps here as well.

Also, please add a section in the readme how to update golang dependencies.

Makefile
```diff
@@ -44,6 +44,13 @@ $(OUT_DIR)/$(BIN)-%:
 
 build: $(OUT_DIR)/$(BIN)
 
+.PHONY: update-go-deps
+update-go-deps:
+	@for m in $$(go list -mod=readonly -m -f '{{ if and (not .Indirect) (not .Main)}}{{.Path}}{{end}}' all); do \
+		go get $$m; \
```
s-urbaniak, line 50: i believe we want to `go get -d`, see https://github.com/prometheus/prometheus/blob/b48a1e9e9bc3575e2f6ef3e7ea8573434288e0bb/Makefile.common#L159-L168

according to `get help get`:
```
The -d flag instructs get not to build or install packages. get will only
update go.mod and download source code needed to build packages.
```

PR #160, opened 2022-02-03
add update-go-deps and perform initial update to k8s v0.23.1

ibihim:


s-urbaniak:
the build still fails, it seems that the setup-go action has different go versions installed on the container, we have to pick the correct one, see https://github.com/actions/setup-go/issues/107.

ibihim:
> the build still fails, it seems that the setup-go action has different go versions installed on the container, we have to pick the correct one, see [actions/setup-go#107](https://github.com/actions/setup-go/issues/107).

I think that the design of the github action could be better (no default to go1.15), but it is actually our fault. We didn't specify the go version for the `make test-e2e` step and so we went with the default of go-1.15

pkg/authn/oidc.go
```diff
@@ -34,12 +36,34 @@ type OIDCConfig struct {
...
+	cacp, err := newCAContentProvider(config.CAFile)
+	if err != nil {
+		return nil, err
+	}
+
 	tokenAuthenticator, err := oidc.New(oidc.Options{
 		IssuerURL:            config.IssuerURL,
 		ClientID:             config.ClientID,
-		CAFile:               config.CAFile,
+		CAContentProvider:    cacp,
```
s-urbaniak, line 66: please use https://github.com/kubernetes/kubernetes/blob/release-1.23/staging/src/k8s.io/apiserver/pkg/server/dynamiccertificates/dynamic_cafile_content.go#L75 instead.

s-urbaniak: using this will give us rotation of CA content for free.

ibihim: @s-urbaniak, PTAL at the latest commit.

s-urbaniak (changes requested):
just one comment regarding inlining

pkg/authn/oidc.go
```diff
@@ -35,11 +38,17 @@ type OIDCConfig struct {
 }
 
 // NewOIDCAuthenticator returns OIDC authenticator
-func NewOIDCAuthenticator(config *OIDCConfig) (authenticator.Request, error) {
+func NewOIDCAuthenticator(ctx context.Context, config *OIDCConfig) (authenticator.Request, error) {
+	dyCA, err := dynamiccertificates.NewDynamicCAContentFromFile("kube-rbac-proxy", config.CAFile)
+	if err != nil {
+		return nil, err
+	}
+	dyCA.Run(1, ctx.Done())
```
s-urbaniak, line 46: on a second thought (sorry, my bad, I didn't catch this earlier): A constructor shouldn't have these strong side-effects. Let's call the Run methods outside of the constructor, this also aligns it with the delegating authenticator:
```
diff --git a/main.go b/main.go
index fba74831..7bf2b711 100644
--- a/main.go
+++ b/main.go
@@ -174,10 +174,12 @@ func main() {
 		defer cancel()
 
-		authenticator, err = authn.NewOIDCAuthenticator(ctx, cfg.auth.Authentication.OIDC)
+		oidcAuthenticator, err := authn.NewOIDCAuthenticator(ctx, cfg.auth.Authentication.OIDC)
 		if err != nil {
 			klog.Fatalf("Failed to instantiate OIDC authenticator: %v", err)
 		}
+		go oidcAuthenticator.Run(1, context.Background().Done())
+		authenticator = oidcAuthenticator
 	} else {
 		//Use Delegating authenticator
 		klog.Infof("Valid token audiences: %s", strings.Join(cfg.auth.Authentication.Token.Audiences, ", "))
diff --git a/pkg/authn/oidc.go b/pkg/authn/oidc.go
index 702f4ae5..bd5498bf 100644
--- a/pkg/authn/oidc.go
+++ b/pkg/authn/oidc.go
@@ -18,11 +18,11 @@ package authn
 
 import (
 	"context"
-
 	"k8s.io/apiserver/pkg/authentication/authenticator"
 	"k8s.io/apiserver/pkg/authentication/request/bearertoken"
 	"k8s.io/apiserver/pkg/server/dynamiccertificates"
 	"k8s.io/apiserver/plugin/pkg/authenticator/token/oidc"
+	"net/http"
 )
 
 // OIDCConfig represents configuration used for JWT request authentication
@@ -37,13 +37,29 @@ type OIDCConfig struct {
 	SupportedSigningAlgs []string
 }
 
+type OIDCAuthenticator struct {
+	dynamicCA     *dynamiccertificates.DynamicFileCAContent
+	authenticator *bearertoken.Authenticator
+}
+
+func (o *OIDCAuthenticator) RunOnce() error {
+	return o.dynamicCA.RunOnce()
+}
+
+func (o *OIDCAuthenticator) Run(workers int, stopCh <-chan struct{}) {
+	o.dynamicCA.Run(workers, stopCh)
+}
+
+func (o *OIDCAuthenticator) AuthenticateRequest(req *http.Request) (*authenticator.Response, bool, error) {
+	return o.authenticator.AuthenticateRequest(req)
+}
+
 // NewOIDCAuthenticator returns OIDC authenticator
-func NewOIDCAuthenticator(ctx context.Context, config *OIDCConfig) (authenticator.Request, error) {
+func NewOIDCAuthenticator(ctx context.Context, config *OIDCConfig) (*OIDCAuthenticator, error) {
 	dyCA, err := dynamiccertificates.NewDynamicCAContentFromFile("kube-rbac-proxy", config.CAFile)
 	if err != nil {
 		return nil, err
 	}
-	dyCA.Run(1, ctx.Done())
 
 	tokenAuthenticator, err := oidc.New(oidc.Options{
 		IssuerURL:            config.IssuerURL,
@@ -59,5 +75,8 @@ func NewOIDCAuthenticator(ctx context.Context, config *OIDCConfig) (authenticato
 		return nil, err
 	}
 
-	return bearertoken.New(tokenAuthenticator), nil
+	return &OIDCAuthenticator{
+		authenticator: bearertoken.New(tokenAuthenticator),
+		dynamicCA:     dyCA,
+	}, nil
 }
```

s-urbaniak (changes requested):

pkg/authn/oidc.go
ibihim: Yes, it is better to handle the control of go routines as far up as we can. Ideally the person who owns the object that runs it. So I completely agree with you on that topic and I am glad you mentioned it. I was just too much in "lets ship it finally"-mode.

liangyuanpeng (approved):

Makefile
```diff
@@ -85,6 +94,13 @@ test-unit:
 test-e2e:
 	go test -timeout 55m -v ./test/e2e/ $(TEST_RUN_ARGS) --kubeconfig=$(KUBECONFIG)
 
+kind-delete-cluster:
+	kind delete cluster --name $(BIN)
+
+kind-create-cluster: kind-delete-cluster
+	kind create cluster --name $(BIN) --config ./test/e2e/kind-config/kind-config.yaml
+	kind load docker-image $(CONTAINER_NAME)
```
s-urbaniak, line 102: nice 👍

s-urbaniak (approved):
lgtm 🎉

s-urbaniak:
@ibihim please squash into meaningful commits, once merged, we can release a new version 👍

s-urbaniak:
all green, merging

s-urbaniak:
thank you @ibihim ! 🎉 now, let's do a release :-)

PR #162, opened 2022-02-17
sig-auth review: comment on code

ibihim:


main.go
```diff
@@ -1,426 +0,0 @@
...
-
-	// TLS flags
-	flagset.StringVar(&cfg.tls.certFile, "tls-cert-file", "", "File containing the default x509 Certificate for HTTPS. (CA cert, if any, concatenated after server cert)")
-	flagset.StringVar(&cfg.tls.keyFile, "tls-private-key-file", "", "File containing the default x509 private key matching --tls-cert-file.")
-	flagset.StringVar(&cfg.tls.minVersion, "tls-min-version", "VersionTLS12", "Minimum TLS version supported. Value must match version names from https://golang.org/pkg/crypto/tls/#pkg-constants.")
-	flagset.StringSliceVar(&cfg.tls.cipherSuites, "tls-cipher-suites", nil, "Comma-separated list of cipher suites for the server. Values are from tls package constants (https://golang.org/pkg/crypto/tls/#pkg-constants). If omitted, the default Go cipher suites will be used")
-	flagset.DurationVar(&cfg.tls.reloadInterval, "tls-reload-interval", time.Minute, "The interval at which to watch for TLS certificate changes, by default set to 1 minute.")
```
deads2k, line 116: minor: k8s.io/apiserver auto-reloads these now if this depended on that.

main.go
```diff
@@ -1,426 +0,0 @@
...
-	flagset.BoolVar(&cfg.auth.Authentication.Header.Enabled, "auth-header-fields-enabled", false, "When set to true, kube-rbac-proxy adds auth-related fields to the headers of http requests sent to the upstream")
-	flagset.StringVar(&cfg.auth.Authentication.Header.UserFieldName, "auth-header-user-field-name", "x-remote-user", "The name of the field inside a http(2) request header to tell the upstream server about the user's name")
-	flagset.StringVar(&cfg.auth.Authentication.Header.GroupsFieldName, "auth-header-groups-field-name", "x-remote-groups", "The name of the field inside a http(2) request header to tell the upstream server about the user's groups")
-	flagset.StringVar(&cfg.auth.Authentication.Header.GroupSeparator, "auth-header-groups-field-separator", "|", "The separator string used for concatenating multiple group names in a groups header field's value")
-	flagset.StringSliceVar(&cfg.auth.Authentication.Token.Audiences, "auth-token-audiences", []string{}, "Comma-separated list of token audiences to accept. By default a token does not have to have any specific audience. It is recommended to set a specific audience.")
```
deads2k, line 124: This default means that a kube-apiserver token would be accepted by this proxy?  This looks like a risky default.

main.go
```diff
@@ -1,426 +0,0 @@
...
-	flagset := pflag.NewFlagSet(os.Args[0], pflag.ExitOnError)
-	flagset.AddGoFlagSet(klogFlags)
-
-	// kube-rbac-proxy flags
-	flagset.StringVar(&cfg.insecureListenAddress, "insecure-listen-address", "", "The address the kube-rbac-proxy HTTP server should listen on.")
-	flagset.StringVar(&cfg.secureListenAddress, "secure-listen-address", "", "The address the kube-rbac-proxy HTTPs server should listen on.")
-	flagset.StringVar(&cfg.upstream, "upstream", "", "The upstream URL to proxy to once requests have successfully been authenticated and authorized.")
-	flagset.BoolVar(&cfg.upstreamForceH2C, "upstream-force-h2c", false, "Force h2c to communiate with the upstream. This is required when the upstream speaks h2c(http/2 cleartext - insecure variant of http/2) only. For example, go-grpc server in the insecure mode, such as helm's tiller w/o TLS, speaks h2c only")
-	flagset.StringVar(&cfg.upstreamCAFile, "upstream-ca-file", "", "The CA the upstream uses for TLS connection. This is required when the upstream uses TLS and its own CA certificate")
-	flagset.StringVar(&configFileName, "config-file", "", "Configuration file to configure kube-rbac-proxy.")
```
deads2k, line 107: This is the authorization config, distinct from the authentication config?

pkg/authz/auth.go
```diff
@@ -1,149 +0,0 @@
...
-			return staticConf == requestVal
-		}
-	}
-
-	userName := ""
-	if a.GetUser() != nil {
-		userName = a.GetUser().GetName()
-	}
-
-	if isAllowed(saConfig.User.Name, userName) &&
```
deads2k, line 117: Allowing an empty username (and really any field) in the file to match all users looks like a recipe for "mispelled a value and now any user can do this!"

TODO: post-acceptance, start a path to have a safe (non-matching) default and migrate with warnings, flags, and forced changes.


explicitly using a "*" instead is safer and doesn't match valid names.

pkg/authz/auth.go
```diff
@@ -1,149 +0,0 @@
...
-		DenyCacheTTL:              30 * time.Second,
-	}
-	return authorizerConfig.New()
-}
-
-type staticAuthorizer struct {
-	config []StaticAuthorizationConfig
-}
-
-func (saConfig StaticAuthorizationConfig) Equal(a authorizer.Attributes) bool {
```
deads2k, line 103: minor: this looks more like `Matches` than `Equal`

pkg/authz/auth.go
```diff
@@ -1,149 +0,0 @@
...
-			return authorizer.DecisionAllow, "found corresponding static auth config", nil
-		}
-	}
-
-	return authorizer.DecisionNoOpinion, "", nil
-}
-
-func NewStaticAuthorizer(config []StaticAuthorizationConfig) (*staticAuthorizer, error) {
-	for _, c := range config {
-		if c.ResourceRequest != (c.Path == "") {
```
deads2k, line 144: post-acceptance, validate this in an explicit "validate user input" stage in addition to having it here.

pkg/authz/auth_test.go
```diff
@@ -1,159 +0,0 @@
...
-
-import (
-	"context"
-	"testing"
-
-	"k8s.io/apiserver/pkg/authentication/user"
-	"k8s.io/apiserver/pkg/authorization/authorizer"
-)
-
-func TestStaticAuthorizer(t *testing.T) {
```
deads2k, line 27: missing wildcard non-resource checks?

deads2k, line 27: missing multiple path non-resource requests.  `/heathz/detailedfailure` for instance.

pkg/authn/delegating.go
```diff
@@ -1,84 +0,0 @@
...
-	}
-
-	var (
-		p   *dynamiccertificates.DynamicFileCAContent
-		err error
-	)
-
-	authenticatorConfig := authenticatorfactory.DelegatingAuthenticatorConfig{
-		Anonymous:               false, // always require authentication
-		CacheTTL:                2 * time.Minute,
```
deads2k, line 48: What are the k8s.io/apiserver defaults

TODO: post-acceptance, path to make defaults match.

pkg/proxy/proxy.go
```diff
@@ -1,267 +0,0 @@
...
-	// config for kube-rbac-proxy
-	Config Config
-}
-
-func new(authenticator authenticator.Request, authorizer authorizer.Authorizer, config Config) *kubeRBACProxy {
-	return &kubeRBACProxy{authenticator, authorizer, newKubeRBACProxyAuthorizerAttributesGetter(config.Authorization), config}
-}
-
-// New creates an authenticator, an authorizer, and a matching authorizer attributes getter compatible with the kube-rbac-proxy
-func New(client clientset.Interface, config Config, authorizer authorizer.Authorizer, authenticator authenticator.Request) (*kubeRBACProxy, error) {
```
deads2k, line 57: error is always nil.

pkg/proxy/proxy.go
```diff
@@ -1,267 +0,0 @@
...
-
-// Config holds proxy authorization and authentication settings
-type Config struct {
-	Authentication *authn.AuthnConfig
-	Authorization  *authz.Config
-}
-
-type kubeRBACProxy struct {
-	// authenticator identifies the user for requests to kube-rbac-proxy
-	authenticator.Request
```
deads2k, line 43: minor: prefer not directly embedding types.

pkg/proxy/proxy.go
```diff
@@ -1,267 +0,0 @@
...
-// If the authn fails, a 401 error is returned. If the authz fails, a 403 error is returned
-func (h *kubeRBACProxy) Handle(w http.ResponseWriter, req *http.Request) bool {
-	ctx := req.Context()
-	if len(h.Config.Authentication.Token.Audiences) > 0 {
-		ctx = authenticator.WithAudiences(ctx, h.Config.Authentication.Token.Audiences)
-		req = req.WithContext(ctx)
-	}
-
-	// Authenticate
-	u, ok, err := h.AuthenticateRequest(req)
```
deads2k, line 71: this does not appear to match kube `filters.WithAuthentication`.  What is different and can we use the upstream filter

pkg/proxy/proxy.go
```diff
@@ -1,267 +0,0 @@
...
-func New(client clientset.Interface, config Config, authorizer authorizer.Authorizer, authenticator authenticator.Request) (*kubeRBACProxy, error) {
-	return new(authenticator, authorizer, config), nil
-}
-
-// Handle authenticates the client and authorizes the request.
-// If the authn fails, a 401 error is returned. If the authz fails, a 403 error is returned
-func (h *kubeRBACProxy) Handle(w http.ResponseWriter, req *http.Request) bool {
-	ctx := req.Context()
-	if len(h.Config.Authentication.Token.Audiences) > 0 {
-		ctx = authenticator.WithAudiences(ctx, h.Config.Authentication.Token.Audiences)
```
deads2k, line 66: not seeing where this is checked yet.  Where is it?  Can we match kube and do it during authentication.

pkg/proxy/proxy.go
```diff
@@ -1,267 +0,0 @@
...
-}
-
-type krpAuthorizerAttributesGetter struct {
-	authzConfig *authz.Config
-}
-
-// GetRequestAttributes populates authorizer attributes for the requests to kube-rbac-proxy.
-func (n krpAuthorizerAttributesGetter) GetRequestAttributes(u user.Info, r *http.Request) []authorizer.Attributes {
-	apiVerb := ""
-	switch r.Method {
```
deads2k, line 130: Falling through with an empty verb isn't what I would expect.  Why does it do this instead of `Unknown_r.Method` or failing?

pkg/proxy/proxy.go
```diff
@@ -1,267 +0,0 @@
...
-			Path:            r.URL.Path,
-		})
-		return allAttrs
-	}
-
-	if n.authzConfig.Rewrites == nil {
-		allAttrs := append(allAttrs, authorizer.AttributesRecord{
-			User:            u,
-			Verb:            apiVerb,
-			Namespace:       n.authzConfig.ResourceAttributes.Namespace,
```
deads2k, line 172: Do these values match what was provided in the --config from back in main.go?

pkg/proxy/proxy.go
```diff
@@ -1,267 +0,0 @@
...
-			params = append(params, ps...)
-		}
-	}
-	if n.authzConfig.Rewrites.ByHTTPHeader != nil && n.authzConfig.Rewrites.ByHTTPHeader.Name != "" {
-		if p := r.Header.Get(n.authzConfig.Rewrites.ByHTTPHeader.Name); p != "" {
-			params = append(params, p)
-		}
-	}
-
-	if len(params) == 0 {
```
deads2k, line 195: This is a case where the configuration means: "forward this request without performing authorization"?  I need to check the config file format, but I'm surprised that the default is "don't protect".  Am I reading this incorrectly?

pkg/proxy/proxy.go
```diff
@@ -1,267 +0,0 @@
...
-		return allAttrs
-	}
-
-	for _, param := range params {
-		attrs := authorizer.AttributesRecord{
-			User:            u,
-			Verb:            apiVerb,
-			Namespace:       templateWithValue(n.authzConfig.ResourceAttributes.Namespace, param),
```
deads2k, line 203: So this means that the same param is passed to every possible resource attribute for golang template substitution?  That means that the header and query param can only logically control a single aspect without differentiating between header and query param?

Making sure I understand and that it's well described.  The behavior is acceptable.

deads2k: Actually, this seems somewhat risky.  If I specify multiple headers and get one to succeed, but the others fail, how can we be assured that the consuming application honors the same header or query param that passed?  

This must be documented as an "if any of these values, controlled by the caller, pass validation, the request is allowed" case.

pkg/proxy/proxy.go
```diff
@@ -1,267 +0,0 @@
...
-	for _, attrs := range allAttrs {
-		// Authorize
-		authorized, reason, err := h.Authorize(ctx, attrs)
-		if err != nil {
-			msg := fmt.Sprintf("Authorization error (user=%s, verb=%s, resource=%s, subresource=%s)", u.User.GetName(), attrs.GetVerb(), attrs.GetResource(), attrs.GetSubresource())
-			klog.Errorf("%s: %s", msg, err)
-			http.Error(w, msg, http.StatusInternalServerError)
-			return false
-		}
-		if authorized != authorizer.DecisionAllow {
```
deads2k, line 100: This changes no-opinion to false instead of falling through, making the behavior dependent on the ordering of headers in a logical map, right?

main.go
```diff
@@ -1,426 +0,0 @@
...
-	flagset.StringVar(&cfg.tls.minVersion, "tls-min-version", "VersionTLS12", "Minimum TLS version supported. Value must match version names from https://golang.org/pkg/crypto/tls/#pkg-constants.")
-	flagset.StringSliceVar(&cfg.tls.cipherSuites, "tls-cipher-suites", nil, "Comma-separated list of cipher suites for the server. Values are from tls package constants (https://golang.org/pkg/crypto/tls/#pkg-constants). If omitted, the default Go cipher suites will be used")
-	flagset.DurationVar(&cfg.tls.reloadInterval, "tls-reload-interval", time.Minute, "The interval at which to watch for TLS certificate changes, by default set to 1 minute.")
-
-	// Auth flags
-	flagset.StringVar(&cfg.auth.Authentication.X509.ClientCAFile, "client-ca-file", "", "If set, any request presenting a client certificate signed by one of the authorities in the client-ca-file is authenticated with an identity corresponding to the CommonName of the client certificate.")
-	flagset.BoolVar(&cfg.auth.Authentication.Header.Enabled, "auth-header-fields-enabled", false, "When set to true, kube-rbac-proxy adds auth-related fields to the headers of http requests sent to the upstream")
```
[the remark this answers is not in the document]
ibihim: The TLS config can be specified [here](https://github.com/brancz/kube-rbac-proxy/blob/master/main.go#L112-L116=).
It is used [here](https://github.com/brancz/kube-rbac-proxy/blob/master/main.go#L299-L344=) to specify [tls.Config](https://pkg.go.dev/crypto/tls#Config).

If no TLS cert / key is provided, is creates a self-signed one.

main.go
ibihim: Yes, the config file is authz only. Authentication is handled by flags. Might have grown organically in time.

The pre-acceptance task is to update documentation in CLI and README, not the uniform handling of configuration options, right?

pkg/proxy/proxy.go
ibihim: The Audiences are used for the [DelegatingAuthenticatorConfig](https://github.com/brancz/kube-rbac-proxy/blob/master/pkg/authn/delegating.go#L46-L51=). This leverages [kubernetes logic](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/plugin/pkg/authenticator/token/webhook/webhook.go#L111=).

pkg/proxy/proxy.go
ibihim: The `authenticator.Request` interface is at the core of [filters.WithAuthentication](https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/staging/src/k8s.io/apiserver/pkg/endpoints/filters/authentication.go#L56=).
The biggest difference is that there are no metrics taken and that the audience check happens in the `DelegatingAuthenticatorConfig` part.

pkg/proxy/proxy.go
ibihim: Yes. The [configfile](https://github.com/brancz/kube-rbac-proxy/blob/master/main.go#L151-L163=) is unmarshaled into the Config type that has the [Namespace](https://github.com/brancz/kube-rbac-proxy/blob/master/pkg/authz/auth.go#L60=) `ResourceAttributes`.

pkg/proxy/proxy.go
ibihim: Well, the logic is not trivial and begs to break on code changes, but it looks like so:

1. We start with [empty allAttr](https://github.com/brancz/kube-rbac-proxy/blob/master/pkg/proxy/proxy.go#L143=).
2. The next two if-blocks are returning.
3. The next two if-blocks fill the  `params`.
4. If `params` is empty, return `allAttrs`, which is by then still empty.
5. And [here](https://github.com/brancz/kube-rbac-proxy/blob/master/pkg/proxy/proxy.go#L83-L84=) we fail, if the returned `allAttrs` is empty.

pkg/proxy/proxy.go
ibihim: We are iterating over all elements of `allAttrs` and all elements need to be authorized. This is `AND` logic. So the ordering doesn't matter.

main.go
ibihim: Yes, this is what it does.

pkg/proxy/proxy.go
ibihim: But only in the particular case of `n.authzConfig.Rewrites == nil`.

pkg/proxy/proxy.go
ibihim: So as far as I understand it, 

1. We can specify values through query (multiple) and through the header (first).
2. Those values get individually applied to the config.
3. Where there are template markers, those get replaced.
4. We append it to the `authorizer.AttributesRecord`.

main.go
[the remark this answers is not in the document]
deads2k: > The TLS config can be specified [here](https://github.com/brancz/kube-rbac-proxy/blob/master/main.go#L112-L116=).
It is used [here](https://github.com/brancz/kube-rbac-proxy/blob/master/main.go#L299-L344=) to specify [tls.Config](https://pkg.go.dev/crypto/tls#Config).

>If no TLS cert / key is provided, is [creates a self-signed one](https://github.com/brancz/kube-rbac-proxy/blob/master/main.go#L300-L314=).

Those are serving certificates, but not client certificates

main.go
deads2k: > The pre-acceptance task is to update documentation in CLI and README, not the uniform handling of configuration options, right?

correct.  Ideally with links to where I can see what format I've got here.  I couldn't tell what this was without reading the code.

pkg/proxy/proxy.go
deads2k: TODO: pre-acceptance, collapse onto upstream for consistency.

pkg/proxy/proxy.go
deads2k: TODO: pre-acceptance closing the empty verb hole

pkg/proxy/proxy.go
deads2k: TODO: pre-acceptance.  headers are map[string][]string.  This code needs to properly handle that.

pkg/proxy/proxy.go
deads2k: TODO: pre-acceptance this should proactively error from this method so that we aren't succeeding by coordination in two unrelated parts.

pkg/proxy/proxy.go
```diff
@@ -1,267 +0,0 @@
...
-			Name:            templateWithValue(n.authzConfig.ResourceAttributes.Name, param),
-			ResourceRequest: true,
-		}
-		allAttrs = append(allAttrs, attrs)
-	}
-	return allAttrs
-}
-
-// DeepCopy of Proxy Configuration
-func (c *Config) DeepCopy() *Config {
```
deads2k, line 217: TODO: pre-acceptance either remove or fuzz test. This is similar to the deep copy of rest.config in k8s.io/client-go to be sure that adding fields doesn't result in non-deepcopiable results or incorrect deep copies.

transport.go
```diff
@@ -1,61 +0,0 @@
...
-		return nil, fmt.Errorf("error reading upstream CA file: %v", err)
-	}
-
-	roots := x509.NewCertPool()
-	if ok := roots.AppendCertsFromPEM([]byte(rootPEM)); !ok {
-		return nil, errors.New("error parsing upstream CA certificate")
-	}
-
-	// http.Transport sourced from go 1.10.7
-	transport := &http.Transport{
```
deads2k, line 46: TODO: pre-acceptance diff with k/k transport defaults

main.go
```diff
@@ -1,426 +0,0 @@
...
-	proxy.Transport = upstreamTransport
-
-	if cfg.upstreamForceH2C {
-		// Force http/2 for connections to the upstream i.e. do not start with HTTP1.1 UPGRADE req to
-		// initialize http/2 session.
-		// See https://github.com/golang/go/issues/14141#issuecomment-219212895 for more context
-		proxy.Transport = &http2.Transport{
-			// Allow http schema. This doesn't automatically disable TLS
-			AllowHTTP: true,
-			// Do disable TLS.
```
deads2k, line 248: TODO: pre-acceptance figure out why it's save to disable TLS.  If it isn't safe to disable TLS (seems questionable for a proxy) stop doing it.

main.go
```diff
@@ -1,426 +0,0 @@
...
-		}
-	}
-
-	mux := http.NewServeMux()
-	mux.Handle("/", http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
-		found := len(cfg.allowPaths) == 0
-		for _, pathAllowed := range cfg.allowPaths {
-			found, err = path.Match(pathAllowed, req.URL.Path)
-			if err != nil {
-				return
```
deads2k, line 262: If we have an error, why not at least return a 500 to the user?

TODO: pre-acceptance either return an error (doesn't have to be the real error) or write a test proving that this produces a error.

main.go
```diff
@@ -1,426 +0,0 @@
...
-			// Do disable TLS.
-			// In combination with the schema check above. We could enforce h2c against the upstream server
-			DialTLS: func(netw, addr string, cfg *tls.Config) (net.Conn, error) {
-				return net.Dial(netw, addr)
-			},
-		}
-	}
-
-	mux := http.NewServeMux()
-	mux.Handle("/", http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
```
deads2k, line 257: I'm not seeing a unit test for this

TODO: pre-acceptance this mux handling method must have have 100% unit test coverage.

main.go
```diff
@@ -1,426 +0,0 @@
...
-		if !found {
-			http.NotFound(w, req)
-			return
-		}
-
-		ignorePathFound := false
-		for _, pathIgnored := range cfg.ignorePaths {
-			ignorePathFound, err = path.Match(pathIgnored, req.URL.Path)
-			if err != nil {
-				return
```
deads2k, line 277: same error comment as above.

main.go
```diff
@@ -1,426 +0,0 @@
...
-		}
-
-		if !ignorePathFound {
-			ok := auth.Handle(w, req)
-			if !ok {
-				return
-			}
-		}
-
-		proxy.ServeHTTP(w, req)
```
deads2k, line 291: Flag for diff against the kube-apiserver upgradeaware proxy.  I'm surprised a different proxy is being used here than the aggregated apiserver proxy.  I remember issues with ugprade awareness.

main.go
```diff
@@ -1,426 +0,0 @@
...
-				cert, err := tls.X509KeyPair(certBytes, keyBytes)
-				if err != nil {
-					klog.Fatalf("Failed to load generated self signed cert and key: %v", err)
-				}
-
-				srv.TLSConfig.Certificates = []tls.Certificate{cert}
-			} else {
-				klog.Info("Reading certificate files")
-				ctx, cancel := context.WithCancel(context.Background())
-				r, err := rbac_proxy_tls.NewCertReloader(cfg.tls.certFile, cfg.tls.keyFile, cfg.tls.reloadInterval)
```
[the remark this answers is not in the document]
deads2k: > note: find this package.

prefer the k8s.io/apiserver package that does this.

main.go
```diff
@@ -1,426 +0,0 @@
...
-			}
-		}
-
-		proxy.ServeHTTP(w, req)
-	}))
-
-	var gr run.Group
-	{
-		if cfg.secureListenAddress != "" {
-			srv := &http.Server{Handler: mux, TLSConfig: &tls.Config{}}
```
deads2k, line 297: most of this actually looks like it wants k8s.io/apiserver ServingOptions for the ports, bind addresses, certs, tls ciphers, etc.

main.go
```diff
@@ -1,426 +0,0 @@
...
-			cipherSuiteIDs, err := k8sapiflag.TLSCipherSuites(cfg.tls.cipherSuites)
-			if err != nil {
-				klog.Fatalf("Failed to convert TLS cipher suite name to ID: %v", err)
-			}
-
-			srv.TLSConfig.CipherSuites = cipherSuiteIDs
-			srv.TLSConfig.MinVersion = version
-			srv.TLSConfig.ClientAuth = tls.RequestClientCert
-
-			if err := http2.ConfigureServer(srv, nil); err != nil {
```
deads2k, line 346: TODO, if this isn't replaced with upstream's serving library, the tls and listener logic here needs review.

main.go
```diff
@@ -1,426 +0,0 @@
...
-			// In combination with the schema check above. We could enforce h2c against the upstream server
-			DialTLS: func(netw, addr string, cfg *tls.Config) (net.Conn, error) {
-				return net.Dial(netw, addr)
-			},
-		}
-	}
-
-	mux := http.NewServeMux()
-	mux.Handle("/", http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
-		found := len(cfg.allowPaths) == 0
```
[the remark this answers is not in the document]
ibihim: I agree. It might be prone to fatal mistakes.

main.go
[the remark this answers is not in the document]
s-urbaniak: yes, agreed, this flag should gain additional client-cert/key parameters. My only historical guess why this is currently as is:

- upstream and kube-rbac-proxy are always running in the same pod
- upstream listens (usually) on localhost only
- hence, a rogue client would have to enter the pod's network namespace to execute rogue requests
- if a malicious client can enter pod's network namespace, it can most likely also enter the pod's mount namespace and introspect configures client certs/keys

anyways, definitely a good and necessary addition 👍

main.go
```diff
@@ -1,426 +0,0 @@
...
-
-	// Add klog flags
-	klogFlags := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
-	klog.InitFlags(klogFlags)
-
-	flagset := pflag.NewFlagSet(os.Args[0], pflag.ExitOnError)
-	flagset.AddGoFlagSet(klogFlags)
-
-	// kube-rbac-proxy flags
-	flagset.StringVar(&cfg.insecureListenAddress, "insecure-listen-address", "", "The address the kube-rbac-proxy HTTP server should listen on.")
```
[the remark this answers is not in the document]
s-urbaniak: historically this was made a separate setting in https://github.com/brancz/kube-rbac-proxy/commit/904be987c272a360c3fcbe2de5a48a91a5fe9714.

I do agree this is dangerous territory. insecure listening is maybe just useful for debugging. I am fine with removing it. We should make it clear though that this is a breaking change.

main.go
s-urbaniak: No, this is for the raw go http proxy server cert/key reload and not dependent on k8s.io/apiserver.

s-urbaniak: However, we can investigate if k8s.io/apiserver proxy code can be reused 👍

main.go
s-urbaniak: @deads2k do you mind to elaborate? I recall clayton also mentioned this during the sig-auth call, but I honestly don't remember 😬

main.go
```diff
@@ -1,426 +0,0 @@
...
-	flagset.StringVar(&cfg.tls.keyFile, "tls-private-key-file", "", "File containing the default x509 private key matching --tls-cert-file.")
-	flagset.StringVar(&cfg.tls.minVersion, "tls-min-version", "VersionTLS12", "Minimum TLS version supported. Value must match version names from https://golang.org/pkg/crypto/tls/#pkg-constants.")
-	flagset.StringSliceVar(&cfg.tls.cipherSuites, "tls-cipher-suites", nil, "Comma-separated list of cipher suites for the server. Values are from tls package constants (https://golang.org/pkg/crypto/tls/#pkg-constants). If omitted, the default Go cipher suites will be used")
-	flagset.DurationVar(&cfg.tls.reloadInterval, "tls-reload-interval", time.Minute, "The interval at which to watch for TLS certificate changes, by default set to 1 minute.")
-
-	// Auth flags
-	flagset.StringVar(&cfg.auth.Authentication.X509.ClientCAFile, "client-ca-file", "", "If set, any request presenting a client certificate signed by one of the authorities in the client-ca-file is authenticated with an identity corresponding to the CommonName of the client certificate.")
```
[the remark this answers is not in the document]
s-urbaniak: @deads2k what would be a sensible in-cluster location?

pkg/proxy/proxy.go
s-urbaniak: didn't find good precedence in k8s which quickly looking but we should investigate.

pkg/proxy/proxy.go
s-urbaniak: let's remove, it seems unused.

main.go
s-urbaniak: @deads2k what is kube's stance on supporting h2c? If kube does not explicitly support it, i think this is fine to be removed.

s-urbaniak: (again, a breaking change with announcement necessary if we decide that path)

main.go
```diff
@@ -1,426 +0,0 @@
...
-
-		proxy.ServeHTTP(w, req)
-	}))
-
-	var gr run.Group
-	{
-		if cfg.secureListenAddress != "" {
-			srv := &http.Server{Handler: mux, TLSConfig: &tls.Config{}}
-
-			if cfg.tls.certFile == "" && cfg.tls.keyFile == "" {
```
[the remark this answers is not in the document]
s-urbaniak: i do agree

main.go
s-urbaniak: the more i think about the more i believe we should strongly consider reusing k8s.io/apiserver serving logic (and options).

pkg/proxy/proxy.go
s-urbaniak: agreed

main.go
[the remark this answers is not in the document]
deads2k: > @deads2k what would be a sensible in-cluster location?

it's stored in a configmap in kube-public namespace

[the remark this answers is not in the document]
deads2k: Post acceptance, but strongly recommended.

main.go
deads2k: @liggitt @enj transport configuration for proxy

main.go
```diff
@@ -1,426 +0,0 @@
...
-	}
-
-	for _, pathIgnored := range cfg.ignorePaths {
-		_, err := path.Match(pathIgnored, "")
-		if err != nil {
-			klog.Fatalf("Failed to verify ignored path: %s", pathIgnored)
-		}
-	}
-
-	proxy := httputil.NewSingleHostReverseProxy(upstreamURL)
```
deads2k, line 238: @enj @liggitt does it proxy how we'd expect?

pkg/proxy/proxy.go
ibihim: If the method is not specified, SAR returns allowed only if the identity is allowed to use all verbs with `"*"`, but we could make it explicit by starting of with `"*"` instead of `""`.

main.go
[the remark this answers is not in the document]
ibihim: @deads2k, we are only focused on removing an insecure listen address for kube-rbac-proxy, but do we want to continue support for an upstream H2C?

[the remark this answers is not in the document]
deads2k: > @deads2k, we are only focused on removing an insecure listen address for kube-rbac-proxy, but do we want to continue support for an upstream H2C?

I'm ok allowing an http backend as long as we are not asserting identity to it.  If identity is asserted, then the connection should be https.

main.go
ibihim: By doing some reading on h2c, the code and history digging, I found that this is meant to be used for proxy to upstream communication and not anyone else.
It seems like some folks wanted to protect tiller (which did h2c) and what not.
In combination with your comment here @deads2k, which I agree on, I would revisit the configuration / logic for identity forwarding in such cases.

main.go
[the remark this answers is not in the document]
ibihim: https://github.com/kubernetes/apiserver/blob/master/pkg/server/dynamiccertificates/dynamic_serving_content.go#L57

pkg/authz/auth.go
```diff
@@ -1,149 +0,0 @@
...
-}
-
-// NewSarAuthorizer creates an authorizer compatible with the kubelet's needs
-func NewSarAuthorizer(client authorizationclient.SubjectAccessReviewInterface) (authorizer.Authorizer, error) {
-	if client == nil {
-		return nil, errors.New("no client provided, cannot use webhook authorization")
-	}
-	authorizerConfig := authorizerfactory.DelegatingAuthorizerConfig{
-		SubjectAccessReviewClient: client,
-		AllowCacheTTL:             5 * time.Minute,
```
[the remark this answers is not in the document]
ibihim: It is 5 * time.Minute: https://github.com/kubernetes/kubernetes/blob/442574f3a7321ff0f34dec9bae67f28651dbabd2/pkg/kubelet/apis/config/v1beta1/defaults.go#L87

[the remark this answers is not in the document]
ibihim: But this doesn't make that much sense to use some kubelet defaults.
This looks more appropriate: `10 * time.Second`.

https://github.com/kubernetes/kubernetes/blob/3ffdfbe286ebcea5d75617da6accaf67f815e0cf/staging/src/k8s.io/apiserver/pkg/server/options/authorization.go#L79-L95

pkg/authn/delegating.go
ibihim: Seems like `10 * time.Second`.

https://github.com/kubernetes/kubernetes/blob/75b19b242c34f7c1d1545e156f802f0a06b90533/staging/src/k8s.io/apiserver/pkg/server/options/authentication.go#L207-L220

pkg/proxy/proxy.go
ibihim: clientset.Interface is also not used. lol

ibihim:
The purpose of this PR was to create the issues. I would close the PR as of now. The issues are still open and reference key discussion of this PR.

PR #170, opened 2022-04-27
initTransport() uses http.DefaultTransport for initialization.

raptorsun:
This PR let initTransport() uses http.DefaultTransport for initialization, avoiding potential discrepancy on the default HTTP transport initial value between kube-rbac-proxy and golang `net/http` package.

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -22,9 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
-	"net"
 	"net/http"
-	"time"
 )
 
 func initTransport(upstreamCAFile string) (http.RoundTripper, error) {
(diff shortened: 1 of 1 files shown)


ibihim:
The default has ForceAttemptHTTP2` set to `True`:`https://github.com/golang/go/blob/690ac4071fa3e07113bf371c9e74394ab54d6749/src/net/http/transport.go#L49.

We don't specify it in the codebase, so it is `False` by default: https://go.dev/play/p/zPVXusVXSUF

raptorsun:
> The default has ForceAttemptHTTP2`set to`True`:`https://github.com/golang/go/blob/690ac4071fa3e07113bf371c9e74394ab54d6749/src/net/http/transport.go#L49.
> 
> We don't specify it in the codebase, so it is `False` by default: https://go.dev/play/p/zPVXusVXSUF

Thank you for the review :)
I corrected it in the new commit.

ibihim:
@raptorsun, I hope you are patient with me, as I come up with those questions just now.

Why do you want to clone the default transport in the first place? So we are not left out of improvements?

Changes done by the golang language developers could introduce regression to our users. Imagine you update software on your cluster by a minor version jump to get the newest CVE fixes. 
And now your cluster behaves slightly different, which causes friction. There is nothing in the changelog that would help you to understand that change. It would be very hard to identify a small change in the timeouts of a default transport in Go 1.18.x that got pulled in through a kube-rbac-proxy minor version bump a couple of versions before.

Maybe I am too cautious, and you have a different point of view that would be help to understand why discrepancies would be bad.

ibihim:
Maybe I am too cautious, I reviewed the apiserver's proxy code and they just take the default: https://github.com/kubernetes/apimachinery/blob/master/pkg/util/proxy/transport.go#L99-L101=.

WDYT? @s-urbaniak

raptorsun:
> @raptorsun, I hope you are patient with me, as I come up with those questions just now.
> 
> Why do you want to clone the default transport in the first place? So we are not left out of improvements?
> 
> Changes done by the golang language developers could introduce regression to our users. Imagine you update software on your cluster by a minor version jump to get the newest CVE fixes. And now your cluster behaves slightly different, which causes friction. There is nothing in the changelog that would help you to understand that change. It would be very hard to identify a small change in the timeouts of a default transport in Go 1.18.x that got pulled in through a kube-rbac-proxy minor version bump a couple of versions before.
> 
> Maybe I am too cautious, and you have a different point of view that would be help to understand why discrepancies would be bad.

Please excuse me for the late reply ^^" 
The reason is that the default value from the package may be safer than the default value assigned to omitted members of the structure by a compiler, especially for eventual new fields, to avoid undesirable behaviors when the package is updated.

PR #171, opened 2022-04-29
check all headers, go version bump

ibihim:
## What

- set go-version on build workflow to 1.17.9
- remove dead code: DeepCopy
- check for all headers, not just the first entry, when creating AttributesRecords

## Why

- go version 1.17.8 and 1.17.9 have security fixes
- unused code is bad code
- don't give the impression that a second header is checked, even though it isn't

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 env:
   QUAY_PATH: quay.io/brancz/kube-rbac-proxy
-  go-version: '1.17.6'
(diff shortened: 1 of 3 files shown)


s-urbaniak:
@ibihim generally, lgtm. Can you please:

- separate the go bump in a separate commit
- elaborate in the commit fixing the headers the following topics:
  - motivation
  - solution strategy

Currently it is not clear what this is fixing, why and how.

s-urbaniak:
lgtm

PR #183, opened 2022-08-04
knopt - github action run - Build kube-rbac-proxy binary during docker build

ibihim:
Signed-off-by: Tomasz Knopik <tknopik@nvidia.com>


ibihim:
Used to trigger tests.

PR #184, opened 2022-08-04
bump: golang.org/x/crypto due to CVE-2022-27191

ibihim:
We are not affected by the CVE, but security scanners push people to
create issues on our repository.


Darnesey:
Thanks for pushing a patch!

s-urbaniak (approved):

PR #185, opened 2022-08-04
bump: emicklei/go-restful due to CVE-2022-1996

ibihim:
We are not affected by the CVE directly or indirectly. CVE scanners push
people to create issues against our repository.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,6 +17,8 @@ require (
 	k8s.io/klog/v2 v2.70.0
 )
 
+replace github.com/emicklei/go-restful => github.com/emicklei/go-restful v2.16.0+incompatible // indirect
+
 require (
 	github.com/NYTimes/gziphandler v1.1.1 // indirect
 	github.com/PuerkitoBio/purell v1.1.1 // indirect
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -17,6 +17,8 @@ require (
 	k8s.io/klog/v2 v2.70.0
 )
 
+replace github.com/emicklei/go-restful => github.com/emicklei/go-restful v2.16.0+incompatible // indirect
```
Darnesey, line 20: Has this version been tested against a trivy scan (or some other security scanner?) I haven't tested to see if `v2.16.0+incompatible` passes the security scan. I know trivy claims the fix version is `v3.8.0`.

Darnesey: Never mind. I see the fix in their source code was cherrypicked into their 2.16.0 version. Thanks!
https://github.com/emicklei/go-restful/blob/v2.16.0/cors_filter.go

Darnesey:
Thanks for pushing a patch!

s-urbaniak (approved):

Issue #169, opened 2022-04-25
Sig-Auth Pre-Acceptance TODOs


ibihim:
- [x] Remove `insecure-listen-address` as it is dangerous to do so with an auth proxy ([link](https://github.com/brancz/kube-rbac-proxy/pull/162#discussion_r809240492))
- [x] Add client certificates for kube-rbac-proxy to upstream ([link](https://github.com/brancz/kube-rbac-proxy/pull/162#discussion_r809243899))
- [ ] Update developer documentation to emphasize the distinction between authn opts and authz configs and show the config format ([link](https://github.com/brancz/kube-rbac-proxy/pull/162#discussion_r809247665)).
- [x] Don't allow attributes requests without http methods ([link](https://github.com/brancz/kube-rbac-proxy/pull/162#discussion_r834566803)).
- [x] Improve logic coordination of in proxy.go to not rely on other parts of the code base to handle empty attributes ([link](https://github.com/brancz/kube-rbac-proxy/pull/162#discussion_r840636813)).
- [x] Check all headers with an `AND` logic ([link](https://github.com/brancz/kube-rbac-proxy/pull/162#discussion_r840631319)).
- [x] Remove unused DeepCopy ([link](https://github.com/brancz/kube-rbac-proxy/pull/162#discussion_r840639502)).
- [x] Diff with `http.Transport` defaults in k/k ([link](https://github.com/brancz/kube-rbac-proxy/pull/162#discussion_r840642346)).
- [ ] H2C topics and disabling TLS ([link](https://github.com/brancz/kube-rbac-proxy/pull/162#discussion_r840644641)).
- [x] Return errors to users, when those happen ([link](https://github.com/brancz/kube-rbac-proxy/pull/162#discussion_r840647634), [link](https://github.com/brancz/kube-rbac-proxy/pull/162#discussion_r840652866)).
- [x] Mux logic needs 100% test coverage and / or improve readability ([link](https://github.com/brancz/kube-rbac-proxy/pull/162#discussion_r840650815), [link](https://github.com/brancz/kube-rbac-proxy/pull/162#discussion_r840651647)).
- [x] Explain what a bool return value on the `kubeRBACProxy` `Handle` means ([link](https://github.com/brancz/kube-rbac-proxy/pull/162#discussion_r840654580)).
- [x] Reuse k8s.io/apiserver serving logic ([link](https://github.com/brancz/kube-rbac-proxy/pull/162#discussion_r840662987)).
[...]

PR #186, opened 2022-08-04
remove insecure listen address

ibihim:
Offering authn / authz to an insecure server is risky.

The benefits vs risks are easier debugging vs potential for an insecure auth-handling-proxy. The risk outweigh the benefits.

[Ref Issue](https://github.com/brancz/kube-rbac-proxy/issues/169)


s-urbaniak:
We can't merge this as is unfortunately, we'd break the CLI API. What we should do instead:

- officially deprecate the CLI setting
- have log warnings saying this is deprecated
- create an issue that outlines the deprecation and wil break once this will be moved to sig-auth

ibihim:
Will be reopened, once we close in on its merge.

ibihim:
Will be merged into v1 branch

PR #190, opened 2022-08-17
[v1] Post acceptance use apiserver cert reloader

ibihim:
# What

Use k8s tls reloader.

# Why

The mantra of the project is to use k8s logic where ever possible.

# Ref

https://github.com/brancz/kube-rbac-proxy/issues/168

PR #195, opened 2022-08-24
WIP: Use SecureServingOptions and DelegatingAuthenticationOptions from new…

xueqzhan:
… apiserver package

This enables in cluster client CA lookup if ClientCA is not configured.

This also enables dynamic tls config update when ClientCA or server certificate change.


main.go
```diff
@@ -58,81 +60,135 @@ type config struct {
...
 
 	// TLS flags
-	flagset.StringVar(&cfg.tls.certFile, "tls-cert-file", "", "File containing the default x509 Certificate for HTTPS. (CA cert, if any, concatenated after server cert)")
-	flagset.StringVar(&cfg.tls.keyFile, "tls-private-key-file", "", "File containing the default x509 private key matching --tls-cert-file.")
-	flagset.StringVar(&cfg.tls.minVersion, "tls-min-version", "VersionTLS12", "Minimum TLS version supported. Value must match version names from https://golang.org/pkg/crypto/tls/#pkg-constants.")
-	flagset.StringSliceVar(&cfg.tls.cipherSuites, "tls-cipher-suites", nil, "Comma-separated list of cipher suites for the server. Values are from tls package constants (https://golang.org/pkg/crypto/tls/#pkg-constants). If omitted, the default Go cipher suites will be used")
-	flagset.DurationVar(&cfg.tls.reloadInterval, "tls-reload-interval", time.Minute, "The interval at which to watch for TLS certificate changes, by default set to 1 minute.")
```
deads2k, line 115: removing this will insta-break callers on upgrade.  You could disconnect the variable so it doesn't drive anything, but leave the flag registered and deprecated with a message indicating that it is now based on fsnotify.

main.go
```diff
@@ -58,81 +60,135 @@ type config struct {
...
-	flagset.StringSliceVar(&cfg.auth.Authentication.Token.Audiences, "auth-token-audiences", []string{}, "Comma-separated list of token audiences to accept. By default a token does not have to have any specific audience. It is recommended to set a specific audience.")
+	flags.StringVar(&c.AuthenticationOptions.ClientCert.ClientCA, "client-ca-file", "", "If set, any request presenting a client certificate signed by one of the authorities in the client-ca-file is authenticated with an identity corresponding to the CommonName of the client certificate.")
+	flags.BoolVar(&c.AuthenticationOptions.SkipInClusterLookup, "authentication-skip-lookup", true, ""+
+		"If false, the authentication-kubeconfig will be used to lookup missing authentication "+
+		"configuration from the cluster.")
+	flags.BoolVar(&c.auth.Authentication.Header.Enabled, "auth-header-fields-enabled", false, "When set to true, kube-rbac-proxy adds auth-related fields to the headers of http requests sent to the upstream")
```
deads2k, line 112: this wiring looks incorrect.  Is this used to determine how to handle requests arriving at the kube-rbac-proxy or used to determine how to send requests to the thing kube-rbac-proxy is sitting in front of?

deads2k: > this wiring looks incorrect. Is this used to determine how to handle requests arriving at the kube-rbac-proxy or used to determine how to send requests to the thing kube-rbac-proxy is sitting in front of?

Oh, it only diffed due to name changes. nm

main.go
```diff
@@ -58,81 +60,135 @@ type config struct {
...
-	configFileName := ""
-
-	// Add klog flags
-	klogFlags := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
-	klog.InitFlags(klogFlags)
+}
 
-	flagset := pflag.NewFlagSet(os.Args[0], pflag.ExitOnError)
-	flagset.AddGoFlagSet(klogFlags)
+func (c *config) AddFlags(flags *pflag.FlagSet) {
```
deads2k, line 86: no vanity renames in this PR.  Don't use flags, use flagset.  Don't use c, use cfg

main.go
```diff
@@ -190,6 +229,15 @@ For more information, please go to https://github.com/brancz/kube-rbac-proxy/iss
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
+	var (
+		secureServingInfo    *server.SecureServingInfo
+		loopbackClientConfig *rest.Config
```
deads2k, line 234: godoc please.  This is weird to see here.  Dead? If so, can you make it up in the method you're calling?

main.go
```diff
@@ -324,76 +364,31 @@ For more information, please go to https://github.com/brancz/kube-rbac-proxy/iss
 	var gr run.Group
 	{
 		if cfg.secureListenAddress != "" {
-			srv := &http.Server{Handler: mux, TLSConfig: &tls.Config{}}
+			shutdownTimeout := time.Duration(60) * time.Second
+			stopCh := make(chan struct{})
+			var stoppedCh <-chan struct{}
```
deads2k, line 369: why is this and listenerStoppedCh scoped out here instead of inside the anonymous func?

test/e2e/basics.go
```diff
@@ -491,12 +491,12 @@ func testIgnorePaths(client kubernetes.Interface) kubetest.TestSuite {
 			Then: kubetest.Actions(
 				kubetest.ClientSucceeds(
 					client,
-					fmt.Sprintf(commandWithoutAuth, "/", 401, 401),
+					fmt.Sprintf(commandWithoutAuth, "/", 403, 403),
```
deads2k, line 494: This probably changed because the delegating kube authenticator allows anonymous.  I think this is change is reasonable and in keeping with other delegated authentication usages, but it's worth calling out to a maintainer here and seeing how to write up a release note about this change and how versions should be numbered.  

The kube delegated authentication (used by kube-controller-manager for instance) doesn't even offer a no-anonymous option.

main.go
xueqzhan: It is used to determine how to send requests from kube-rbac-proxy to upstream. Therefore this is independent of the apiserver request header authentication flags. Thus I left this as it is.

README.md
```diff
@@ -120,6 +120,13 @@ To run tests locally, you need to have [kind](https://kind.sigs.k8s.io/) install
 
 The command to execute the tests is: `VERSION=local make clean container kind-create-cluster test`.
 
+If building from mac, you need to split the command into two parts:
```
s-urbaniak, line 123: pleas introduce a `## Building` paragraph.

main.go
s-urbaniak: agreed

main.go
s-urbaniak: yes, we cannot do this on this branch, this will break existing workloads.

test/e2e/basics.go
s-urbaniak: cc @ibihim for ^

main.go
xueqzhan: Will put it back.

main.go
xueqzhan: I am testing removing this now. I think we might not need the loopback support here. Will update based on result.

test/e2e/basics.go
xueqzhan: It will be interesting to know if anyone is using this without authz. I thought this is rbac proxy. So authz should be its main functionality. But maybe people use it differently?

README.md
xueqzhan: Sorry, I don't mean building here. There is no differences in building for this PR. I meant to add a paragraph for testing e2e. I will change the wording.

main.go
xueqzhan: As it turns out, loopback adds self-signed server certificate. This is needed when no certificate is provided from command line options.

main.go
```diff
@@ -62,60 +64,59 @@ type config struct {
...
+	if cfg.ServingOptions == nil || cfg.AuthenticationOptions == nil {
+		return
+	}
 
 	// kube-rbac-proxy flags
 	flagset.StringVar(&cfg.insecureListenAddress, "insecure-listen-address", "", "The address the kube-rbac-proxy HTTP server should listen on.")
 	flagset.StringVar(&cfg.secureListenAddress, "secure-listen-address", "", "The address the kube-rbac-proxy HTTPs server should listen on.")
 	flagset.StringVar(&cfg.upstream, "upstream", "", "The upstream URL to proxy to once requests have successfully been authenticated and authorized.")
-	flagset.BoolVar(&cfg.upstreamForceH2C, "upstream-force-h2c", false, "Force h2c to communiate with the upstream. This is required when the upstream speaks h2c(http/2 cleartext - insecure variant of http/2) only. For example, go-grpc server in the insecure mode, such as helm's tiller w/o TLS, speaks h2c only")
+	flagset.BoolVar(&cfg.upstreamForceH2C, "upstream-force-h2c", false, "Force h2c to communiate with the upstream. This is required when the upstream speaks h2c(http/2 cleartext - insecure variant of http/2) only. For example, go-grpc server in the insecure mode, such as helm's tiller w/c TLS, speaks h2c only")
```
deads2k, line 102: w/o was correct.

main.go
```diff
@@ -62,60 +64,59 @@ type config struct {
...
+	flagset.StringSliceVar(&cfg.ServingOptions.CipherSuites, "tls-cipher-suites", nil, "Comma-separated list of cipher suites for the server. Values are from tls package constants (https://golang.org/pkg/crypto/tls/#pkg-constants). If omitted, the default Go cipher suites will be used")
 	flagset.DurationVar(&cfg.tls.reloadInterval, "tls-reload-interval", time.Minute, "The interval at which to watch for TLS certificate changes, by default set to 1 minute.")
 
 	// Auth flags
-	flagset.StringVar(&cfg.auth.Authentication.X509.ClientCAFile, "client-ca-file", "", "If set, any request presenting a client certificate signed by one of the authorities in the client-ca-file is authenticated with an identity corresponding to the CommonName of the client certificate.")
+	flagset.StringVar(&cfg.AuthenticationOptions.ClientCert.ClientCA, "client-ca-file", "", "If set, any request presenting a client certificate signed by one of the authorities in the client-ca-file is authenticated with an identity corresponding to the CommonName of the client certificate.")
+	flagset.BoolVar(&cfg.AuthenticationOptions.SkipInClusterLookup, "authentication-skip-lookup", true, ""+
```
deads2k, line 117: if the KIND cluster has a client CA generated for it and a functional CSR client API, then you can
1. set just this flag to --true, 
2. don't set the --client-ca-file
3. create a client cert/key pair
4. then use cluster-admin creds to approve the CSR
5. the kube-controller-manager should sign using a signing key that matches the in-cluster client-ca
6. get the signed cert from CSR status
7. then you can use your new client cert/key again the kube-rbac-server

test/e2e/basics.go
deads2k: if authz is optional in kube-rbac-proxy params, then you need a way to disable anonymous auth and the spot is here: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/options/authentication.go#L286

main.go
xueqzhan: No idea how this got here. :(

test/e2e/basics.go
xueqzhan: If I [read this ](https://github.com/brancz/kube-rbac-proxy/blob/fc1ca4f969941340a8adb66932bd64dc5773d37a/pkg/proxy/proxy.go#L83)correctly, there is no way to run kube-rbac-proxy without authz. @ibihim can you confirm?

main.go
xueqzhan: Done

main.go
```diff
@@ -133,6 +134,69 @@ func main() {
...
+		cfg.ServingOptions.BindAddress = net.ParseIP(host)
+		cfg.ServingOptions.BindPort = port
+
+	}
+
+	if err = cfg.ServingOptions.ApplyTo(secureServingInfo, loopbackClientConfig); err != nil {
+		klog.Fatalf("Failed to apply serving options: %v", err)
+	}
+	if err = cfg.AuthenticationOptions.ApplyTo(authenticationInfo, *secureServingInfo, nil); err != nil {
+		klog.Fatalf("Failed to apply authentication options: %v", err)
```
ibihim, line 184: Nit: we could use `%w` insted of `%v`.

main.go
```diff
@@ -163,24 +227,16 @@ For more information, please go to https://github.com/brancz/kube-rbac-proxy/iss
...
-	if configFileName != "" {
-		klog.Infof("Reading config file: %s", configFileName)
-		b, err := os.ReadFile(configFileName)
-		if err != nil {
-			klog.Fatalf("Failed to read resource-attribute file: %v", err)
-		}
-
-		configfile := configfile{}
+Flag --tls-reload-interval will be removed from future releases. It is here for
+backward compatibility. New tls reload mechanism takes care of this automatically.
```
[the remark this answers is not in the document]
ibihim: @s-urbaniak, we talked about it and we would like to such a change in a v1 release?

ibihim (commented):
Not completely through yet, but wanted to share already.

Great effort. Keep going.

main.go
```diff
@@ -133,6 +134,69 @@ func main() {
 
 	//Kubeconfig flag
 	flagset.StringVar(&cfg.kubeconfigLocation, "kubeconfig", "", "Path to a kubeconfig file, specifying how to connect to the API server. If unset, in-cluster configuration will be used")
+}
+
+type configfile struct {
+	AuthorizationConfig *authz.Config `json:"authorization,omitempty"`
+}
+
+func (cfg *config) ApplyTo(secureServingInfo **server.SecureServingInfo, loopbackClientConfig **rest.Config, authenticationInfo *server.AuthenticationInfo) error {
```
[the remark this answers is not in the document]
xueqzhan: Yes, this is here because of https://github.com/kubernetes/apiserver/blob/edc3b6f28ef9b20877b9b7f9b4cb7fbf004abdc0/pkg/server/options/serving_with_loopback.go#L39.

main.go
```diff
@@ -62,60 +64,59 @@ type config struct {
...
-	configFileName := ""
+}
 
-	// Add klog flags
-	klogFlags := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
-	klog.InitFlags(klogFlags)
+// tlsConfig is no longer used. TLS configuration is now managed by SecureServingOptions
+// This is here to keep tls-reload-interval for backward compatibility
+type tlsConfig struct {
+	reloadInterval time.Duration
```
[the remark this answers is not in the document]
xueqzhan: Do you have the link? This will be a blocker if it is a valid objection. :(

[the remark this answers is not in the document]
deads2k: we cannot remove cli flags without prior deprecation.

Keep the flag so invocations don't fail. Once you switch to upstream kube fsnotify, you can disconnect the flag from the variable and print a warning that the reload is driven by fsnotify, not by polling.

ref: https://github.com/brancz/kube-rbac-proxy/pull/190#discussion_r949014647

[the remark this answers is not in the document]
xueqzhan: I see it now. I thought the objection was for using k8s cert reloader. 

With this PR, I am keeping the reloadInterval flag and dump a warning about the potential future removal of the flag. Will that be good enough?

stlaz:
This was delivered in https://github.com/brancz/kube-rbac-proxy/pull/215 instead. @ibihim you can close this PR.

ibihim:
Ok, cool. Let me close it. If I've done a mistake, feel free to reopen it again.

PR #206, opened 2022-11-01
feat: Add healthz path

nabokihms:
Signed-off-by: m.nabokikh <maksim.nabokikh@flant.com>

## Motivation
There is no way to ensure that the proxy itself is safe and sound. One option is to use health checks of an upstream, but in the case of upstream unavailability, both upstream and kube-rbac-proxy will be restarted.

This PR adds an option to optionally configure a health path.

## Additional info
I also fixed some typos.


ibihim:
Sorry, I was busy. I will try to take a look this week.

PR #211, opened 2022-11-29
add client cert auth for upstream connections

stlaz:
This PR exposes the options to specify client cert/key pair for https://github.com/brancz/kube-rbac-proxy/issues/169

/assign @ibihim

### DIFF
--- a/pkg/authn/config.go
+++ b/pkg/authn/config.go
@@ -40,7 +40,9 @@ type AuthnConfig struct {
 
 // X509Config holds public client certificate used for authentication requests if specified
 type X509Config struct {
-	ClientCAFile string
+	ClientCAFile              string
+	UpstreamClientCertificate string
+	UpstreamClientKey         string
 }
 
 // TokenConfig holds configuration as to how token authentication is to be done
--- a/transport.go
+++ b/transport.go
@@ -27,18 +27,26 @@ import (
 	"time"
 )
 
-func initTransport(upstreamCAFile string) (http.RoundTripper, error) {
+func initTransport(upstreamCAFile, upstreamClientCertPath, upstreamClientKeyPath string) (http.RoundTripper, error) {
 	if upstreamCAFile == "" {
 		return http.DefaultTransport, nil
 	}
 
-	rootPEM, err := os.ReadFile(upstreamCAFile)
+	upstreamCAPEM, err := os.ReadFile(upstreamCAFile)
 	if err != nil {
-		return nil, fmt.Errorf("error reading upstream CA file: %v", err)
+		return nil, fmt.Errorf("error reading upstream CA file: %w", err)
 	}
 
-	roots := x509.NewCertPool()
-	if ok := roots.AppendCertsFromPEM([]byte(rootPEM)); !ok {
+	var certKeyPair tls.Certificate
+	if len(upstreamClientCertPath) > 0 {
+		certKeyPair, err = tls.LoadX509KeyPair(upstreamClientCertPath, upstreamClientKeyPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read upstream client cert/key: %w", err)
+		}
+	}
+
+	upstreamCAPool := x509.NewCertPool()
+	if ok := upstreamCAPool.AppendCertsFromPEM([]byte(upstreamCAPEM)); !ok {
 		return nil, errors.New("error parsing upstream CA certificate")
 	}
 
@@ -54,7 +62,13 @@ func initTransport(upstreamCAFile string) (http.RoundTripper, error) {
 		IdleConnTimeout:       90 * time.Second,
 		TLSHandshakeTimeout:   10 * time.Second,
 		ExpectContinueTimeout: 1 * time.Second,
-		TLSClientConfig:       &tls.Config{RootCAs: roots},
+		TLSClientConfig: &tls.Config{
+			RootCAs: upstreamCAPool,
+		},
+	}
+
+	if certKeyPair.Certificate != nil {
+		transport.TLSClientConfig.Certificates = []tls.Certificate{certKeyPair}
 	}
 
 	return transport, nil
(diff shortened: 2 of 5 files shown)


pkg/authn/config.go
```diff
@@ -40,7 +40,9 @@ type AuthnConfig struct {
 
 // X509Config holds public client certificate used for authentication requests if specified
 type X509Config struct {
-	ClientCAFile string
+	ClientCAFile              string
+	UpstreamClientCertificate string
+	UpstreamClientKey         string
```
ibihim, line 45: `ClientCAFile` is in `authn.AuthnConfig` as it is used to authenticated downstream clients by their CN.

Wouldn't it be better to put it into the `main.tlsConfig`?

transport.go
```diff
@@ -27,18 +27,26 @@ import (
...
 )
 
-func initTransport(upstreamCAFile string) (http.RoundTripper, error) {
+func initTransport(upstreamCAFile, upstreamClientCertPath, upstreamClientKeyPath string) (http.RoundTripper, error) {
 	if upstreamCAFile == "" {
 		return http.DefaultTransport, nil
 	}
 
-	rootPEM, err := os.ReadFile(upstreamCAFile)
+	upstreamCAPEM, err := os.ReadFile(upstreamCAFile)
```
ibihim, line 35: Is there a [apiserver / dynamiccertifcates](https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/apiserver/pkg/server/dynamiccertificates) solution we could leverage here?

I think for the aggregated certs, something like [this](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kube-aggregator/pkg/apiserver/apiserver.go#L264) is used.

transport_test.go
```diff
@@ -42,3 +57,143 @@ func TestInitTransportWithCustomCA(t *testing.T) {
...
+		if !isCorrectRequest {
+			t.Errorf("the server did not receive the expected request")
+		}
+	case <-time.NewTimer(1 * time.Second).C:
+		t.Fatalf("the server did not receive any requests in the required time")
+	}
+
+}
+
+func generateClientCert(t *testing.T) ([]byte, []byte, *x509.CertPool) {
```
ibihim, line 158: It would be better to return an error and handle the error case with one `t.Fatal` in the test itself.

transport_test.go
```diff
@@ -42,3 +57,143 @@ func TestInitTransportWithCustomCA(t *testing.T) {
 		t.Error("expected root CA to be set, got nil")
 	}
 }
+
+type testHTTPHandler struct {
```
ibihim, line 61: Could we use [httptest.Server](https://pkg.go.dev/net/http/httptest#Server) instead?

transport_test.go
```diff
@@ -42,3 +57,143 @@ func TestInitTransportWithCustomCA(t *testing.T) {
...
+
+func (h *testHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	if len(req.TLS.PeerCertificates) > 0 {
+		h.gotCorrectRequest <- true
+	} else {
+		reqDump, _ := httputil.DumpRequest(req, false)
+		h.t.Logf("got request without client certificates:\n%s", reqDump)
+		h.t.Logf("TLS config: %#v", req.TLS)
+
+		h.gotCorrectRequest <- false
```
ibihim, line 74: Couldn't we signal the result by using HTTP Status Codes instead and put the logs into the body?

transport_test.go
```diff
@@ -42,3 +57,143 @@ func TestInitTransportWithCustomCA(t *testing.T) {
...
+	}
+}
+
+func TestInitTransportWithClientCertAuth(t *testing.T) {
```
ibihim, line 78: I think it is amazing to have a test like that, but it makes me wonder if we are testing our code or the golang libraries.

ibihim (commented):
Great contribution! I am very grateful for any help wrt the kubernetes pre-acceptance issue.

transport.go
stlaz: Yes. We should do that in a follow-up that would bring us closer to the apiserver serving structures.

transport_test.go
stlaz: I did not find a way how to modify its TLS configuration properly. Even if we did that, you still need to supply your own handler.

transport_test.go
stlaz: You still need to wait for the server to acknowledge that it got the request and processed it.

We could send the error response in a body.

transport_test.go
stlaz: I got something very similar wrong in the past, and it was an issue with the code.

pkg/authn/config.go
stlaz: indeed, that makes sense

transport.go
stlaz: My other PR - https://github.com/brancz/kube-rbac-proxy/pull/210 - is there to make the transition to upstream option structures smoother

transport_test.go
stlaz: I was so stuck on my solution that I almost forgot how normal client-server communication works 

I'll rework this, then.

ibihim (commented):
Do you think you would be able to create an e2e test for that setup? It would have a great verification-that-it-works and documentation purpose.

transport.go
stlaz: I propose to address it along with `Reuse k8s.io/apiserver serving logic` which will likely bring these dynamic cert providers for other bundles, too.

stlaz:
> Do you think you would be able to create an e2e test for that setup? It would have a great verification-that-it-works and documentation purpose.

We would need to find an image that allows testing client cert authentication. I am not currently aware of any, I'll open an issue so that this can be done as a follow-up.

PR #216, opened 2022-12-13
use a different port for proxy-specific endpoints

stlaz:
Using a port as a namespace for endpoints specific to the proxy is a bit more practical than using a path/subpath for each specific endpoint/endpoints as this introduced approach is less prone to upstream endpoint changes.

/assign @ibihim 
/cc @nabokihms

### DIFF
--- a/cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
+++ b/cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
@@ -29,6 +29,7 @@ import (
 	"os"
 	"os/signal"
 	"path"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
@@ -127,7 +128,7 @@ type configfile struct {
 type completedProxyRunOptions struct {
 	insecureListenAddress string // DEPRECATED
 	secureListenAddress   string
-	healthzPath           string
+	proxyEndpointsPort    int
 
 	upstreamURL      *url.URL
 	upstreamForceH2C bool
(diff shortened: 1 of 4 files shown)


cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
```diff
@@ -413,24 +410,59 @@ func Run(cfg *completedProxyRunOptions) error {
...
+
+					klog.Info("Listening securely on %v for proxy endpoints", endpointsAddr)
+					tlsListener := tls.NewListener(proxyListener, srv.TLSConfig)
+					return proxyEndpointsSrv.Serve(tlsListener)
+				}, func(err error) {
+					if err := proxyEndpointsSrv.Shutdown(context.Background()); err != nil {
+						klog.Errorf("failed to gracefully shutdown proxy endpoints server: %w", err)
+					}
+				})
+			}
```
ibihim, line 465: This is pretty cool, but could we maybe move that out of Run and into a dedicated function? Don't know if it might be too much effort wrt all the groupRun and other thingies.

ibihim (commented):
Would it be possible to add a simple e2e test?

ibihim (approved):

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
ibihim: ok, lets do it in one rush.

PR #218, opened 2022-12-15
*: cut v0.14.0 release

ibihim:


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 0.14.0 / 2022-12-15
+
```
stlaz, line 2: Do you want to outline the new feature branch and the future plans with the repository here?

ibihim: A changelog is a changelog. Can't remember to have seen any plans in it somewhere.

BUT I could add that the README.md was changed with information about that, but wouldn't go as far as to give it a section.
This would keep the logical integrity of the CHANGELOG and we would increase the surface of our users to the winds of change.

WDYT?

stlaz: makes sense :+1:

PR #219, opened 2022-12-23
Accept Multi Values in One Header When Using byHttpHeader

Kirchen99:
### Description
This PR extends the feature in https://github.com/brancz/kube-rbac-proxy/pull/104. With a "separator" in the config the header value can be split.

e.g.

```
"authorization":
  "resourceAttributes":
    "apiVersion": "metrics.k8s.io/v1beta1"
    "namespace": "{{ .Value }}"
    "resource": "pods"
  "rewrites":
    "byHttpHeader":
      "name": "X-Scope-OrgID"
      "separator": "|"
```

### Motivation
We also put kube-rbac-proxy in front of Loki (https://grafana.com/docs/loki/latest/). And Loki 2.6 enabled multi-tenant queries: https://grafana.com/blog/2022/07/27/grafana-loki-2.6-release/

The required format from loki is like this:
```
curl -H 'X-Scope-OrgID:Tenant1|Tenant2|Tenant3' \
  -G -s "http://localhost:3100/loki/api/v1/query" \
  --data-urlencode \
  'query=sum(rate({job="varlogs"}[10m])) by (level)' | jq
```

### DIFF
--- a/pkg/authz/auth.go
+++ b/pkg/authz/auth.go
@@ -53,6 +53,7 @@ type QueryParameterRewriteConfig struct {
 // be used to rewrite a SubjectAccessReview on a given request.
 type HTTPHeaderRewriteConfig struct {
 	Name string `json:"name,omitempty"`
+	Sep  string `json:"separator,omitempty"`
 }
 
 // ResourceAttributes describes attributes available for resource request authorization
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"net/http"
 	"net/textproto"
+	"strings"
 	"text/template"
 
 	"github.com/brancz/kube-rbac-proxy/pkg/authn"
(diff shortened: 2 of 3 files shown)


dbluxo:
Hi @ibihim, we would also find this feature super helpful, could you maybe look at this PR again?

pkg/authz/auth.go
```diff
@@ -53,6 +53,7 @@ type QueryParameterRewriteConfig struct {
 // be used to rewrite a SubjectAccessReview on a given request.
 type HTTPHeaderRewriteConfig struct {
 	Name string `json:"name,omitempty"`
+	Sep  string `json:"separator,omitempty"`
```
sathieu, line 56:
```suggestion
	Separator  string `json:"separator,omitempty"`
```

NB: I'm not a reviewer here, just adding a suggestion.

stlaz:
Let's stick to https://datatracker.ietf.org/doc/html/rfc8941#name-structured-data-types and use the standard HTTP separator for the list/dict structured types.

PR #231, opened 2023-04-05
v0.15.0-candidate - go.*: update dependencies

ibihim:


Makefile
```diff
@@ -94,8 +94,11 @@ test-unit:
 test-e2e:
 	go test -timeout 55m -v ./test/e2e/ $(TEST_RUN_ARGS) --kubeconfig=$(KUBECONFIG)
 
-test-local:
-	@echo 'run: VERSION=local make clean container kind-create-cluster test'
+test-local-setup: clean $(OUT_DIR)/$(BIN)-$(GOOS)-$(GOARCH) Dockerfile
```
utkarshn-veritas, line 97: As there is no test executed under this target, instead of test-local-setup should it be image-build?

PR #237, opened 2023-05-12
propagate GOOS to docker container build in test local setup

nickytd:
This PR propagates the GOOS environment variable to the docker build fixing the issue on macOS arm64 platform.


ibihim:
It is not enough, but a step forward.

Issue #238, opened 2023-05-30
Sig-Auth Pre-Acceptance 2nd Review


ibihim:
# What

Issues collected during the second round of review with @enj.

# Why

kube-rbac-proxy should become a project owned by Kubernetes.

# Issues

`cmd/kube-rbac-proxy/app/kube-rbac-proxy.go`:
- [x] #317
- [x] (2) Rename to `ApplyTo` to conform naming convention [link](https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-8796db11a6b137095413f7a7109dd01393e63d425454fc04647b98d2b8c0115cR81).
- [x] (3) Replace `func` with `cobra.Args` [link](https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-8796db11a6b137095413f7a7109dd01393e63d425454fc04647b98d2b8c0115cR106).
- [x] (4) Run should run with a final config version. Complete should finish the necessary changes and those shouldn't happen anymore in `Run`. [link](https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-8796db11a6b137095413f7a7109dd01393e63d425454fc04647b98d2b8c0115cR166).
- [x] (5) Use context from Signals. [link](https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-8796db11a6b137095413f7a7109dd01393e63d425454fc04647b98d2b8c0115cR171).
- [x] #318
- [x] #316
- [x] #319
- [x] (9) H2C is fine as plaintext if listening to 127.0.0.1 / ::1 or UDS [link](https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-8796db11a6b137095413f7a7109dd01393e63d425454fc04647b98d2b8c0115cR198).
- [x] (10) Describe what an empty request info factory means [link](https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-8796db11a6b137095413f7a7109dd01393e63d425454fc04647b98d2b8c0115cR217).
- [x] (11) Use safe wait group from k/k instead of run.Group [link](https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-8796db11a6b137095413f7a7109dd01393e63d425454fc04647b98d2b8c0115cR223)
[...]

PR #229, opened 2023-03-30
[WIP] Mo review

enj:

PR #243, opened 2023-06-07
Minor improvements in cmd/kube-rbac-proxy

liouk:
This PR addresses some of the issues from #238; in particular, the ones listed in [this comment](https://github.com/brancz/kube-rbac-proxy/issues/238#issuecomment-1571662143).

In particular:
- (3) Replace func with cobra.Args [link](https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-8796db11a6b137095413f7a7109dd01393e63d425454fc04647b98d2b8c0115cR106): see 71da752a034bed8df7d4b9da410bd61de4c9f2d1
- (4) Run should run with a final config version. Complete should finish the necessary changes and those shouldn't happen anymore in Run. [link](https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-8796db11a6b137095413f7a7109dd01393e63d425454fc04647b98d2b8c0115cR166): see b410b9693c63eef2181097da627df19c1c88a338
- (5) Use context from Signals. [link](https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-8796db11a6b137095413f7a7109dd01393e63d425454fc04647b98d2b8c0115cR171): see 0c10f9caf132baa2cee91219cb4c8ba7de6f4b24
- (10) Describe what an empty request info factory means [link](https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-8796db11a6b137095413f7a7109dd01393e63d425454fc04647b98d2b8c0115cR217): see c68149625dd689704e8b51b83c3a16fa2b5be1d3
- (11) Use safe wait group from k/k instead of run.Group [link](https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-8796db11a6b137095413f7a7109dd01393e63d425454fc04647b98d2b8c0115cR223)
- (13) Comment why we have two listeners on different ports [link](https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-8796db11a6b137095413f7a7109dd01393e63d425454fc04647b98d2b8c0115cR232): see b410b9693c63eef2181097da627df19c1c88a338
- (15) Add serverconfig.SetupSignalContext [link](https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-8796db11a6b137095413f7a7109dd01393e63d425454fc04647b98d2b8c0115cR242-R243): see 0c10f9caf132baa2cee91219cb4c8ba7de6f4b24
- (17) Authorizers should only be added, when used [link](https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-8796db11a6b137095413f7a7109dd01393e63d425454fc04647b98d2b8c0115cL317-R329): see a24945c11cd8fef7f01e56924a91281b6211fe1e
- (18) Verify that Rewrite Attributes is definitely turned off if not configured [link](https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-8796db11a6b137095413f7a7109dd01393e63d425454fc04647b98d2b8c0115cR314): see a24945c11cd8fef7f01e56924a91281b6211fe1e

### DIFF
--- a/cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
+++ b/cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
@@ -90,7 +90,13 @@ that can perform RBAC authorization against the Kubernetes API using SubjectAcce
 				return utilerrors.NewAggregate(errs)
 			}
 
-			return Run(completedOptions)
+			// create the KubeRBACProxyConfig based on the completed options
+			proxyCfg, err := completedOptions.ProxyConfig()
+			if err != nil {
+				return err
+			}
+
+			return Run(proxyCfg)
 		},
 		Args: cobra.NoArgs,
 	}
@@ -125,6 +131,40 @@ func (o *completedProxyRunOptions) Validate() []error {
 	return errs
 }
 
+func (opts *completedProxyRunOptions) ProxyConfig() (*server.KubeRBACProxyConfig, error) {
+	proxyConfig := server.NewConfig()
+	if err := opts.SecureServing.ApplyTo(&proxyConfig.SecureServing); err != nil {
+		return nil, err
+	}
+
+	if opts.ProxySecureServing != nil {
+		if err := opts.ProxySecureServing.ApplyTo(&proxyConfig.KubeRBACProxyInfo.ProxyEndpointsSecureServing); err != nil {
+			return nil, err
+		}
+	}
+	if err := opts.DelegatingAuthentication.ApplyTo(
+		proxyConfig.DelegatingAuthentication,
+		proxyConfig.SecureServing,
+		nil,
+	); err != nil {
+		return nil, err
+	}
+
+	if err := opts.DelegatingAuthorization.ApplyTo(proxyConfig.DelegatingAuthorization); err != nil {
+		return nil, err
+	}
+
+	if err := opts.ProxyOptions.ApplyTo(proxyConfig.KubeRBACProxyInfo, proxyConfig.DelegatingAuthentication); err != nil {
+		return nil, err
+	}
+
+	if err := opts.OIDCOptions.ApplyTo(proxyConfig.KubeRBACProxyInfo); err != nil {
+		return nil, err
+	}
+
+	return proxyConfig, nil
+}
+
 // Complete sets defaults for the ProxyRunOptions.
 // Should be called after the flags are parsed.
 func Complete(o *options.ProxyRunOptions) (*completedProxyRunOptions, error) {
@@ -152,11 +192,7 @@ func Complete(o *options.ProxyRunOptions) (*completedProxyRunOptions, error) {
 	return completed, nil
 }
 
-func Run(opts *completedProxyRunOptions) error {
-	cfg, err := createKubeRBACProxyConfig(opts)
-	if err != nil {
-		return err
-	}
+func Run(cfg *server.KubeRBACProxyConfig) error {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
@@ -234,40 +270,6 @@ func Run(opts *completedProxyRunOptions) error {
 	}
 
 	return nil
-}
-
-func createKubeRBACProxyConfig(opts *completedProxyRunOptions) (*server.KubeRBACProxyConfig, error) {
-	proxyConfig := server.NewConfig()
-	if err := opts.SecureServing.ApplyTo(&proxyConfig.SecureServing); err != nil {
-		return nil, err
-	}
-
-	if opts.ProxySecureServing != nil {
-		if err := opts.ProxySecureServing.ApplyTo(&proxyConfig.KubeRBACProxyInfo.ProxyEndpointsSecureServing); err != nil {
-			return nil, err
-		}
-	}
-	if err := opts.DelegatingAuthentication.ApplyTo(
-		proxyConfig.DelegatingAuthentication,
-		proxyConfig.SecureServing,
-		nil,
-	); err != nil {
-		return nil, err
-	}
-
-	if err := opts.DelegatingAuthorization.ApplyTo(proxyConfig.DelegatingAuthorization); err != nil {
-		return nil, err
-	}
-
-	if err := opts.ProxyOptions.ApplyTo(proxyConfig.KubeRBACProxyInfo, proxyConfig.DelegatingAuthentication); err != nil {
-		return nil, err
-	}
-
-	if err := opts.OIDCOptions.ApplyTo(proxyConfig.KubeRBACProxyInfo); err != nil {
-		return nil, err
-	}
-
-	return proxyConfig, nil
 }
 
 func secureServerRunner(
--- a/cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
+++ b/cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
@@ -236,6 +236,7 @@ func Run(cfg *server.KubeRBACProxyConfig) error {
 	handler := identityheaders.WithAuthHeaders(proxy, cfg.KubeRBACProxyInfo.UpstreamHeaders)
 	handler = kubefilters.WithAuthorization(handler, authz, scheme.Codecs)
 	handler = kubefilters.WithAuthentication(handler, authenticator, http.HandlerFunc(filters.UnauthorizedHandler), cfg.DelegatingAuthentication.APIAudiences)
+	// passing an empty RequestInfoFactory results in attaching a non-resource RequestInfo to the context
 	handler = kubefilters.WithRequestInfo(handler, &request.RequestInfoFactory{})
 	handler = rewrite.WithKubeRBACProxyParamsHandler(handler, cfg.KubeRBACProxyInfo.Authorization.RewriteAttributesConfig)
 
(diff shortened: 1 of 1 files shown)


cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
```diff
@@ -207,6 +198,7 @@ func Run(opts *completedProxyRunOptions) error {
 	handler := identityheaders.WithAuthHeaders(proxy, cfg.KubeRBACProxyInfo.UpstreamHeaders)
 	handler = kubefilters.WithAuthorization(handler, authz, scheme.Codecs)
 	handler = kubefilters.WithAuthentication(handler, authenticator, http.HandlerFunc(filters.UnauthorizedHandler), cfg.DelegatingAuthentication.APIAudiences)
+	// passing an empty RequestInfoFactory results in attaching a non-resource RequestInfo to the context
```
liouk, line 201: @ibihim I'm not sure if I'm missing any other implication of passing an empty `RequestInfoFactory`, please advise :)

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
```diff
@@ -303,25 +287,26 @@ func secureServerRunner(
...
 	}
 
-	var authz authorizer.Authorizer = rewrite.NewRewritingAuthorizer(
-		union.New(
-			staticAuthorizer,
-			delegatedAuthz.Authorizer,
-		),
-		krbInfo.Authorization.RewriteAttributesConfig,
-	)
+	if krpInfo.Authorization.RewriteAttributesConfig != nil {
```
liouk, line 301: @ibihim I've added a unit test to verify that when not configured explicitly, `RewriteAttributesConfig` is indeed nil; this is the case because [unmarshalling a yaml config file](https://github.com/brancz/kube-rbac-proxy/blob/sig-auth-acceptance/cmd/kube-rbac-proxy/app/options/proxyoptions.go#L147) that doesn't set `RewriteAttributesConfig` will set the field to the zero value which is nil.

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
```diff
@@ -303,25 +287,26 @@ func secureServerRunner(
 	return runner, interrupter
 }
 
-func setupAuthorizer(krbInfo *server.KubeRBACProxyInfo, delegatedAuthz *serverconfig.AuthorizationInfo) (authorizer.Authorizer, error) {
-	staticAuthorizer, err := static.NewStaticAuthorizer(krbInfo.Authorization.Static)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create static authorizer: %w", err)
+func setupAuthorizer(krpInfo *server.KubeRBACProxyInfo, delegatedAuthz *serverconfig.AuthorizationInfo) (authorizer.Authorizer, error) {
```
liouk, line 290: Not sure if applicable, but I renamed `krbInfo` to `krpInfo` to use the initials.

liouk:
Added (11) to the list of addressed issues; converted to Draft while this is WIP.

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
ibihim: Thanks for the clean-up. I assume it is a typo (someone had "kube RBac proxy" in mind?).

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
```diff
@@ -277,14 +264,17 @@ func createKubeRBACProxyConfig(opts *completedProxyRunOptions) (*server.KubeRBAC
 	return proxyConfig, nil
 }
 
+type runnerFunc func() error
+type interrupterFunc func(error)
```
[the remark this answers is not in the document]
liouk: To be honest I considered that, but I still think that keeping it in the same file as the rest is probably clean enough, because:

- it's not code that is/will be used elsewhere within the project, as it's very specific to starting up the KRP
- it's a short piece of code (total 42 loc: typedefs, `secureServerRunner()` and `addToSafeWaitGroup()`), and mainly wrapper code; at the same time the file itself isn't very large (335 loc currently)
- based on the current structure, I'm not sure whether there's a better place, as there are not a lot of dirs within `cmd/kube-rbac-proxy/` and based on the first point it's not very fitting for `pkg/` IMHO

We could always refactor at a later point when we feel that the file is growing too much.

As a side note, this could be simplified a bit more if we dropped the typedefs for the two funcs; they just offer a tiny bit of clarity, but not much since everything is used around in the same spots. What do you think?

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
```diff
@@ -132,17 +122,18 @@ func (o *completedProxyRunOptions) Validate() []error {
 	return errs
 }
 
-// Complete sets defaults for the ProxyRunOptions.
+// Complete sets defaults for the ProxyRunOptions and creates a server.KubeRBACProxyConfig
+// based on the completed options.
 // Should be called after the flags are parsed.
-func Complete(o *options.ProxyRunOptions) (*completedProxyRunOptions, error) {
+func Complete(o *options.ProxyRunOptions) (*completedProxyRunOptions, *server.KubeRBACProxyConfig, error) {
```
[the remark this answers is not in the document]
liouk: This is a great idea; done in 6245f7d5207ca2befcfa4bcb34669b2b9bd09753

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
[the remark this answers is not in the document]
ibihim: Let it leave as is... but 335 loc is not long? 😄 Maybe it is me and I am odd, but I love small files where I can jump in between, rather than scrolling throw long texts.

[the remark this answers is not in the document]
liouk: Well, based on what I've seen so far in our projects I'd say 335 might even be below average :sweat_smile:

I have a suggestion if we want to reduce loc further: we can go with what you've suggested and split runner/interrupter to another file, and we can also split `completedProxyRunOptions` and `Complete()` to another file too. This will cut it down to approx. 200 loc :slightly_smiling_face: WDYT?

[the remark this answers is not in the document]
ibihim: hahaha, yeah. In general in the cloud-space, 3000 LOC seems to be the magical threshold and not 300 😛 

Lets keep it as is. We might want to revisit the whole logic at some point at all.

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
```diff
@@ -294,36 +283,52 @@ func secureServerRunner(
...
 	}
 
-	var authz authorizer.Authorizer = rewrite.NewRewritingAuthorizer(
-		union.New(
-			staticAuthorizer,
-			delegatedAuthz.Authorizer,
-		),
-		krbInfo.Authorization.RewriteAttributesConfig,
-	)
+	if krpInfo.Authorization.RewriteAttributesConfig != nil {
```
ibihim, line 306: I mentioned this to you in Slack, but this introduces a regression. Before doing that change, we would need to do the following:
- What happens to requests, if this piece of code is not run every time?
  - Most probably we won't mask the authorizer.AttirbutesRecord to be a non-resource-request (and some more params).
- IF we don't  do that, what would happen to our requests? Would they default to resource requests? This would most probably not work.

liouk: As we discussed, I dropped commit a24945c11cd8fef7f01e56924a91281b6211fe1e from this PR as we need to tackle this separately and differently.

stlaz:
LGTM, leaving final ack to @ibihim

liouk:
@ibihim after our discussions I've opted for a simple approach to handling the wait groups -- PTAL. I've dropped `secureServerRunner()` as I don't think it's necessary as per your suggestion, and put everything in `Run()`.

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
```diff
@@ -207,100 +233,63 @@ func Run(opts *completedProxyRunOptions) error {
...
-
-	return nil
-}
-
-func createKubeRBACProxyConfig(opts *completedProxyRunOptions) (*server.KubeRBACProxyConfig, error) {
-	proxyConfig := server.NewConfig()
-	if err := opts.SecureServing.ApplyTo(&proxyConfig.SecureServing); err != nil {
-		return nil, err
+	if err := wg.Add(1); err != nil {
+		cancel()
```
ibihim, line 248: Could you move that into line 242 like so: `defer cancel()`?

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
```diff
@@ -207,100 +233,63 @@ func Run(opts *completedProxyRunOptions) error {
...
 
-	if opts.ProxySecureServing != nil {
-		if err := opts.ProxySecureServing.ApplyTo(&proxyConfig.KubeRBACProxyInfo.ProxyEndpointsSecureServing); err != nil {
-			return nil, err
+	go func() {
+		defer wg.Done()
+		stoppedCh, listenerStoppedCh, err := cfg.SecureServing.Serve(mux, 10*time.Second, serverCtx.Done())
+		if err != nil {
+			klog.Errorf("%v", err)
+			cancel()
```
ibihim, line 257: Nit / Not a change request:

We could move it to `defer wg.Done()` with a `defer cancel()`.

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
```diff
@@ -207,100 +233,63 @@ func Run(opts *completedProxyRunOptions) error {
...
-		<-listenerStoppedCh
-		<-stoppedCh
-		return err
-	}
+		go func() {
+			defer wg.Done()
+			stoppedCh, listenerStoppedCh, err := cfg.KubeRBACProxyInfo.ProxyEndpointsSecureServing.Serve(proxyEndpointsMux, 10*time.Second, serverCtx.Done())
+			if err != nil {
+				klog.Errorf("%v", err)
+				cancel()
```
ibihim, line 281: Nit / Not a change request:

We could move it to defer wg.Done() with a defer cancel().

ibihim (commented):
1 comment and 2 nits.

ibihim (approved):
Nice. Thanks for the patience. Looking forward for more PRs 😄

PR #250, opened 2023-07-11
p/a/path: fail on broken pattern early, add tests

ibihim:
# What

- Change path-authorizer to act like upstream (only postfix patterns).
- Fail on broken pattern.
- Add unit tests for Ignore-Path-Case

# Why

- Upstream doesn't use complex patter matching from the standard library. It is prone to misconfiguration.
- If a broken pattern is given, it behaved like a pattern match.
- There were no unit tests for the Ignore-Path-Case.

### DIFF
--- a/cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
+++ b/cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
@@ -298,14 +298,20 @@ func setupAuthorizer(krbInfo *server.KubeRBACProxyInfo, delegatedAuthz *serverco
 	// pathAuthorizer is running before any other authorizer.
 	// It works outside of the default authorizers.
 	var pathAuthorizer authorizer.Authorizer
-
-	// AllowPaths denies access to paths that are not listed.
-	// IgnorePaths doesn't auth(n/z) paths that are listed.
+	var err error
+	// AllowPaths are the only paths that are not denied.
+	// IgnorePaths bypass all authorization checks.
 	switch {
 	case len(krbInfo.AllowPaths) > 0:
-		pathAuthorizer = path.NewAllowPathAuthorizer(krbInfo.AllowPaths)
+		pathAuthorizer, err = path.NewAllowedPathsAuthorizer(krbInfo.AllowPaths)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create allow path authorizer: %w", err)
+		}
 	case len(krbInfo.IgnorePaths) > 0:
-		pathAuthorizer = path.NewAlwaysAllowPathAuthorizer(krbInfo.IgnorePaths)
+		pathAuthorizer, err = path.NewPassthroughAuthorizer(krbInfo.IgnorePaths)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create ignore path authorizer: %w", err)
+		}
 	}
 
 	// Delegated authorizers are running after the pathAuthorizer
--- a/pkg/authorization/path/path.go
+++ b/pkg/authorization/path/path.go
@@ -18,71 +18,41 @@ package path
 
 import (
 	"context"
-	"path"
-	"strings"
+	"fmt"
 
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apiserver/pkg/authorization/authorizer"
+	pathauthorizer "k8s.io/apiserver/pkg/authorization/path"
 )
 
-type pathAuthorizer struct {
-	matchDecision, noMatchDecision authorizer.Decision
+// NewAllowedPathsAuthorizer returns an authorizer that allows requests to
+// `allowPaths` through with `NoOpinion` and denies all others. This ensures that
+// subsequent authorizers in a union are still called on the requests that pass through
+// this authorizer.
+// The provided paths can include simple glob patterns.
+func NewAllowedPathsAuthorizer(allowPaths []string) (authorizer.Authorizer, error) {
+	delegatedPathAuthorizer, err := pathauthorizer.NewAuthorizer(allowPaths)
+	if err != nil {
+		return nil, fmt.Errorf("error creating path authorizer: %v", err)
+	}
 
-	paths        sets.String
-	pathPatterns []string
-}
+	return authorizer.AuthorizerFunc(func(ctx context.Context, attr authorizer.Attributes) (authorizer.Decision, string, error) {
+		decision, reason, err := delegatedPathAuthorizer.Authorize(ctx, attr)
 
-func newPathAuthorizer(onMatch, onNoMatch authorizer.Decision, inputPaths []string) *pathAuthorizer {
-	var patterns []string
-	paths := sets.NewString() // faster than trying to match every pattern every time
-	for _, p := range inputPaths {
-		p = strings.TrimPrefix(p, "/")
-		if len(p) == 0 {
-			// matches "/"
-			paths.Insert(p)
-			continue
+		// There is a match on the path, so we have no opinion and let subsequent
+		// authorizers in a union decide.
+		if err == nil && decision == authorizer.DecisionAllow {
+			return authorizer.DecisionNoOpinion, reason, nil
 		}
-		if strings.ContainsRune(p, '*') {
-			patterns = append(patterns, p)
-		} else {
-			paths.Insert(p)
-		}
-	}
 
-	return &pathAuthorizer{
-		matchDecision:   onMatch,
-		noMatchDecision: onNoMatch,
-		paths:           paths,
-		pathPatterns:    patterns,
-	}
+		return authorizer.DecisionDeny, fmt.Sprintf("NOT(%s)", reason), err
+	}), nil
 }
 
-func (a *pathAuthorizer) Authorize(ctx context.Context, attr authorizer.Attributes) (authorizer.Decision, string, error) {
-	pth := strings.TrimPrefix(attr.GetPath(), "/")
-	if a.paths.Has(pth) {
-		return a.matchDecision, "", nil
-	}
-
-	for _, pattern := range a.pathPatterns {
-		if found, err := path.Match(pattern, pth); err != nil {
-			return authorizer.DecisionNoOpinion, "Error", err
-		} else if found {
-			return a.matchDecision, "", nil
-		}
-	}
-
-	return a.noMatchDecision, "", nil
-}
-
-func NewAllowPathAuthorizer(allowPaths []string) authorizer.Authorizer {
-	if len(allowPaths) == 0 {
-		return authorizer.AuthorizerFunc(func(context.Context, authorizer.Attributes) (authorizer.Decision, string, error) {
-			return authorizer.DecisionNoOpinion, "", nil
-		})
-	}
-	return newPathAuthorizer(authorizer.DecisionNoOpinion, authorizer.DecisionDeny, allowPaths)
-}
-
-func NewAlwaysAllowPathAuthorizer(alwaysAllowPaths []string) authorizer.Authorizer {
-	return newPathAuthorizer(authorizer.DecisionAllow, authorizer.DecisionNoOpinion, alwaysAllowPaths)
+// NewPassthroughAuthorizer returns an authorizer that allows on matches for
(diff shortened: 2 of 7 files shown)


s-urbaniak (approved):

pkg/authorization/path/path.go
```diff
@@ -18,71 +18,46 @@ package path
...
 
-func NewAlwaysAllowPathAuthorizer(alwaysAllowPaths []string) authorizer.Authorizer {
-	return newPathAuthorizer(authorizer.DecisionAllow, authorizer.DecisionNoOpinion, alwaysAllowPaths)
+// NewPassthroughAuthorizer returns an authorizer that allows on matches for
+// the given paths and has no opinion on all others. This allows skipping
+// subsequent authorizers in a union, effectively passing through the given
+// paths.
+// The given paths can include simple glob patterns.
+func NewPassthroughAuthorizer(ignorePaths []string) (authorizer.Authorizer, error) {
+	return pathauthorizer.NewAuthorizer(ignorePaths)
```
stlaz, line 62: this cuts down on functionality

stlaz (changes requested):

pkg/authorization/path/path.go
ibihim: Yes, it does. It is based on your original concept. I wasn't sure if it is a good idea to drop that functionality, but @s-urbaniak said it is fine.

pkg/authorization/rewrite/rewrite.go
```diff
@@ -43,39 +44,58 @@ var _ authorizer.Authorizer = &rewritingAuthorizer{}
...
+	// |-------|----------|
+	// | bar   | allow    |
+	// | baz   | error    |
+	// the result of the union will be allow.
+	//
+	// This would not be an issue in a simple setup as the client defines the
+	// query params and could drop the baz value.
+	// BUT the templating variant is only useful if the upstream service
+	// interprets the query params as well.
+	// AND therefore upstream trusts in the verification of the query params.
```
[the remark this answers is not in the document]
ibihim: The issue is, that the use case is pretty edgy. Usually you bar those feature against huge `unsafe-feature-x` flags 😄
In sum it means that we can't accept the default behavior due to the rewrite use case.

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
```diff
@@ -338,11 +344,12 @@ func setupAuthorizer(krbInfo *server.KubeRBACProxyInfo, delegatedAuthz *serverco
 	}
 
 	rewritingAuthorizer := rewrite.NewRewritingAuthorizer(
-		union.New(delegated...),
+		union.New(delegated...), // union is fine here as static shouldn't fail, if kas fails subsequently, we are fine.
```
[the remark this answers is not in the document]
ibihim: An error is logged, but it doesn't `Deny`. Therefore we need to verify for each flow, if a `NoOpinion` is fine on `error`.

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
```diff
@@ -298,14 +298,20 @@ func setupAuthorizer(krbInfo *server.KubeRBACProxyInfo, delegatedAuthz *serverco
 	// pathAuthorizer is running before any other authorizer.
 	// It works outside of the default authorizers.
 	var pathAuthorizer authorizer.Authorizer
-
-	// AllowPaths denies access to paths that are not listed.
-	// IgnorePaths doesn't auth(n/z) paths that are listed.
+	var err error
+	// AllowPaths are the only paths that are not denied.
+	// IgnorePaths are paths that will get authorized.
```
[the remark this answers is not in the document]
ibihim: Yes, that phrasing sounds good.

pkg/authorization/rewrite/rewrite.go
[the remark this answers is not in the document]
ibihim: In sum it means that we can't accept the default behavior due to the rewrite use case and the comment highlight an example which emphasize this.

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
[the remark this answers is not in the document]
stlaz:
```suggestion
	// IgnorePaths bypass all authorization checks.
```

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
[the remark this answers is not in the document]
stlaz: What I mean is - I don't understand the comment. Why shouldn't union be fine?

pkg/authorization/rewrite/rewrite.go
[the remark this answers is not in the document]
stlaz: I see what this is trying to say.

Union finishes on allow/deny. The original code checked _all_ are allowed, whereas with this change, it checks whether a single combination is allowed.

This change therefore looks like a bug?

stlaz (changes requested):

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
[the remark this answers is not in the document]
ibihim: The default behavior in KRP is more strict then what union provides us.

KRP fails on errors, while union doesn't. This would be a regression.

pkg/authorization/rewrite/rewrite.go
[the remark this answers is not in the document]
ibihim: Oh, sorry. I think you lack the context:

We can't do that:
```go
return union.New(authorizers...).Authorize(ctx, nil)
```

We need to handle the error case like so:
```go
if err != nil {
		return authorizer.DecisionDeny, "error while authorizing", err
}
```.

The reasoning in the comments above is: why must not use just a union.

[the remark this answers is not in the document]
Comment (stlaz) [reply pkg/authorization/rewrite/rewrite.go:86]: That's different from what I'm trying to say.

You cannot use union at all. The union authorizer stops on the first allow. The previous code would only allow if and only if _all_ `proxyAttrs` would allow.

File: docs/path.md [hunk 5]
```diff
@@ -0,0 +1,45 @@
+# Allow Path and Ignore Path
+
+## Description
+
+Paths that are not provided by `--allow-path` will be rejected.
+Paths that are provided by `--ignore-path` won't be rejected.
+It is only possible to define one of both.
```
stlaz, line 7: I think this is a bit confusing. How about:

```suggestion
If `--allow-paths` is set, only the requests to these paths will be considered for proxying, the rest will be rejected.
If `--ignore-paths` is set, requests to these paths will be passed upstream without any authorization.

Only one of `--allow-paths`/`--ignore-paths` is allowed to be set at a time.
```

Note that the README says `allow/ignore-paths` whereas here you're describing it as `allow/ignore-path`

docs/path.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Change
```
stlaz, line 9: Should we have a separate section for changes to previous kube-rbac-versions?

docs/path.md
```diff
@@ -0,0 +1,45 @@
...
+
+Previously **infix regex** worked.
```
stlaz, line 11:
```suggestion
Previously, it was possible to configure an **infix wildcard** but that will now fail.
```

docs/path.md
```diff
@@ -0,0 +1,45 @@
...
+Now **infix regex** will cause an error on server startup.
+kube-rbac-proxy won't start with an **infix regex**.
+E.g. `--allow-path="/api/v1/*/values"` will cause kube-rbac-proxy fail to start.
+E.g. `--ignore-path="/api/v1/*/values"` will cause kube-rbac-proxy fail to start.
+
+Previously `*` would count as a wildcard in between `/`.
```
stlaz, line 17: Wasn't it a single-path-segment wildcard?

docs/path.md
```diff
@@ -0,0 +1,45 @@
...
+Now the wildcard is matching any string, even if it contains `/`.
+E.g. `--allow-path="/api/v1/*"` would have rejected `/api/v1/label/values`, not it evaluated.
+E.g. `--ignore-path="/api/v1/*"` would have evaluated `/api/v1/label/values`, now it is passed through. 
+
+Only `*` is allowed. Previously everything that [Go's standard library path.Match](https://pkg.go.dev/path#Match) allowed worked.
```
stlaz, line 22:
```suggestion
Only a single `*` is allowed. Previously everything that [Go's standard library path.Match](https://pkg.go.dev/path#Match) allowed worked.
```

pkg/authorization/path/path.go
```diff
@@ -18,71 +18,46 @@ package path
...
-			return a.matchDecision, "", nil
-		}
+	// There is a match on the path, so we have no opinion and let subsequent
+	// authorizers in a union decide.
+	if err == nil && decision == authorizer.DecisionAllow {
+		return authorizer.DecisionNoOpinion, reason, nil
 	}
 
-	return a.noMatchDecision, "", nil
+	return authorizer.DecisionDeny, fmt.Sprintf("NOT(%s)", reason), err
```
stlaz, line 40: We don't need a new type for the allowedPathsAuthorizer - inline as
```go
// pseudocode
func NewAllowedPathsAuthorizer(allowPaths []string) (authorizer.Authorizer, error) {
   pathAuthorizer = NewAuthorizer()
   return authorizer.AuthorizerFunc(func() {
        // Authorize() logic here
    })
}
```

pkg/authorization/path/path.go
```diff
@@ -18,71 +18,46 @@ package path
...
 
-func NewAllowPathAuthorizer(allowPaths []string) authorizer.Authorizer {
-	if len(allowPaths) == 0 {
-		return authorizer.AuthorizerFunc(func(context.Context, authorizer.Attributes) (authorizer.Decision, string, error) {
-			return authorizer.DecisionNoOpinion, "", nil
-		})
+// NewAllowedPathsAuthorizer returns an authorizer that has no opinion on
+// matches for the given paths and denies all others. No opinion ensures that
+// subsequent authorizers in a union are still called. The given paths can
+// include simple glob patterns.
```
stlaz, line 46:
```suggestion
// NewAllowedPathsAuthorizer returns an authorizer that allows requests to
// `allowPaths` through with `NoOpinion` and denies all others. This ensures that
// subsequent authorizers in a union are still called on the requests that pass through
// this authorizer.
// The provided paths can include simple glob patterns.
```
What's simple glob patterns?

docs/path.md
ibihim: Do you mean in addition to the CHANGELOG.md?

docs/path.md
```diff
@@ -0,0 +1,44 @@
...
+
+Previously, it was possible to configure an **infix wildcard** but that will now fail.
+Now **infix regex** will cause an error on server startup.
+kube-rbac-proxy won't start with an **infix regex**.
+E.g. `--allow-path="/api/v1/*/values"` will cause kube-rbac-proxy fail to start.
+E.g. `--ignore-path="/api/v1/*/values"` will cause kube-rbac-proxy fail to start.
+
+Previously `*` would count as a single-path-segment wildcard ([path.Match](https://pkg.go.dev/path#Match)).
+Now the wildcard is matching any string, even if it contains `/`.
+E.g. `--allow-path="/api/v1/*"` would have rejected `/api/v1/label/values`, not it evaluated.
```
stlaz, line 20: `not it evaluated` <- that sentence appears to be missing something

docs/path.md
```diff
@@ -0,0 +1,44 @@
...
+E.g. `--ignore-path="/api/v1/*"` would have evaluated `/api/v1/label/values`, now it is passed through.
```
stlaz, line 21:
```suggestion
E.g. `--ignore-path="/api/v1/*"` would have matched `/api/v1/label/values`, now it is passed through. 
```

docs/path.md
```diff
@@ -0,0 +1,44 @@
+# Allow Path and Ignore Path
```
stlaz, line 1: you may want to create a separate `docs/migration` folder for these files

PR #253, opened 2023-09-05
version: set version flag values properly

ibihim:
# What

Set version values during build time.

# Why

Currently we don't set the appropriate version variables and print `Kubernetes v0.0.0-master+$Format:%H$` on `--version`-flag.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -71,7 +71,7 @@ jobs:
         go-version: ${{ env.go-version }}
     - name: Create container & run tests
       run: |
-        VERSION=local make container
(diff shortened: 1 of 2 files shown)


stlaz:
Why are we printing kubernetes version instead of a version of the binary?

Please provide the output of `--version` before this fix and after.

ibihim:
Before

```bash
./kube-rbac-proxy --version                                                                                                                                                                                                                                                         
Kubernetes v0.0.0-master+$Format:%H$
```

After

```bash
./kube-rbac-proxy --version                                                                                                                                                                                                                                                      
kube-rbac-proxy v0.14.2
```

PR #256, opened 2023-09-14
Fixing publish.sh by overriding local tags with `--force`

jhongturney:
adding --force to git fetch --tags to override local without being rejected


ibihim:
This doesn't fix the issue. I was able to overcome the clobber, but now it complains about another amd64 tag. I will continue the investigation on the morning.

PR #257, opened 2023-09-14
Update Makefile

jhongturney:
Would another if/else condition work to drop the tag for amd64? (cc @ibihim)


ibihim:
Nah, I made a mistake in renaming the variables in the Makefile, I need to fix it and update the release. Sorry for any inconveniences.

PR #260, opened 2023-09-23
Fallback to the trusted host ca bundle in the oidc case

### LINKED ISSUE #259: Failing to instantiate OIDC authenticator when --oidc-ca-file is not specified
Hi, 

According to the description of [oidc-ca-file](https://github.com/brancz/kube-rbac-proxy/blob/3ed1e58004ea4bcb493d22addec89aa50d4a10e1/cmd/kube-rbac-proxy/app/options/options.go#L111) parameter, is optional. When it is not specified it shall fallback to the `/etc/ssl/certs/ca-certificates.crt`. The latter is part of the distro-less base image. However the current implementation, using the `k8s.io/apiserver/pkg/server/dynamiccertificates` https://github.com/kubernetes/apiserver/blob/v0.28.2/pkg/server/dynamiccertificates/dynamic_cafile_content.go#L76,  returns an error instead of fallback
```
E0920 14:13:37.375979       1 run.go:74] "command failed" err="failed to instantiate OIDC authenticator: missing filename for ca bundle"
```
The simplest possible fix would be to just set the default as described in the documentation
```
// NewOIDCAuthenticator returns OIDC authenticator
func NewOIDCAuthenticator(config *OIDCConfig) (*OIDCAuthenticator, error) {
	if config.CAFile == "" {
		//If the CA file is not defined, it shall default to the host system trusted ca bundle
		config.CAFile="/etc/ssl/certs/ca-certificates.crt"
	}
	if err != nil {
		return nil, err
	}
...
```
Here is  a PR in case you confirm the problem and agree with the solution proposal.

Thanks
Niki

nickytd:
# What

Attempt to use the trusted system ca bundle in the when such is not provided, following the description of the `--oidc-ca-file` configuration parameter.

# Why

Current behavior emits an error when `--oidc-ca-file` is not provided. Fixes #259

### DIFF
--- a/pkg/authn/oidc.go
+++ b/pkg/authn/oidc.go
@@ -31,12 +31,20 @@ type OIDCAuthenticator struct {
 	requestAuthenticator authenticator.Request
 }
 
+const (
+	trustedHostCABundle = "/etc/ssl/certs/ca-certificates.crt"
+)
+
 var (
 	_ (authenticator.Request) = (*OIDCAuthenticator)(nil)
 )
 
 // NewOIDCAuthenticator returns OIDC authenticator
 func NewOIDCAuthenticator(config *OIDCConfig) (*OIDCAuthenticator, error) {
+	if config.CAFile == "" {
+		//If the CA file is empty, it shall default to the host system trusted ca bundle
+		config.CAFile = trustedHostCABundle
+	}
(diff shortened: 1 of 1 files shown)


ibihim:
Closed with the context of the isssue.

PR #265, opened 2023-10-19
proposal for http2: enable to disable, limit http2 resources

ibihim:
# What

- close HTTP/2 connections on unauthenticated users
- set max streams per connection to 100
- set max frame size and stream buffer to 256kb.
- enable to disable HTTP/2 completely

# Why

- Golang fixes might be not enough [link](https://github.com/kubernetes/kubernetes/pull/121120)

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,9 @@ Kube-rbac-proxy flags:
       --auth-token-audiences strings                Comma-separated list of token audiences to accept. By default a token does not have to have any specific audience. It is recommended to set a specific audience.
       --client-ca-file string                       If set, any request presenting a client certificate signed by one of the authorities in the client-ca-file is authenticated with an identity corresponding to the CommonName of the client certificate.
       --config-file string                          Configuration file to configure kube-rbac-proxy.
+      --http2-disable                               Disable HTTP/2 support
+      --http2-max-concurrent-streams uint32         The maximum number of concurrent streams per HTTP/2 connection. (default 100)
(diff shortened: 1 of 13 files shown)


pkg/filters/auth.go
```diff
@@ -53,6 +54,20 @@ func WithAuthentication(
...
+		// Do not allow unauthenticated clients to keep these
+		// connections open (i.e. basically degrade them to the
+		// performance of http1 with keep-alive disabled).
+		//
+		// https://github.com/kubernetes/apiserver/blob/master/pkg/endpoints/filters/authentication.go#L107C1-L117C4
+		if req.ProtoMajor == 2 && isAnonymousUser(res.User) {
+			// limit this connection to just this request,
+			// and then send a GOAWAY and tear down the TCP connection
+			// https://github.com/golang/net/commit/97aa3a539ec716117a9d15a4659a911f50d13c3c
+			w.Header().Set("Connection", "close")
```
s-urbaniak, line 68: aren't we supposed to return here? you are continuing the handler chain.

ibihim: This closes the TCP connection. The inflight-request will be handled and then the client needs to re-do the TCP handshake, as far as I understood it.

It actually decreases the http2 speed for unauthenticated users to worse then http/1.1 as they need to go through the TCP handshake process.

s-urbaniak (approved):
lgtm

PR #277, opened 2024-01-26
validate options: check for secure connection for h2c and identity headers

ibihim:
## What

Check if mTLS or loopback address is provided on h2c or identity headers.

## Why

h2c is unencrypted http/2 and we don't want to allow it outside of local connections.
request headers functionality is setting confidential in the headers, we want this to happen only through mTLS. We can't guarantee that upstream verifies client certificates.


cmd/kube-rbac-proxy/app/options/proxyoptions.go
```diff
@@ -17,10 +17,16 @@ limitations under the License.
 package options
```
stlaz, line 17: Nit: use capital letters at the beginning of a sentence, even in commit messages.

cmd/kube-rbac-proxy/app/options/proxyoptions.go
```diff
@@ -57,6 +64,7 @@ type ProxyOptions struct {
 func (o *ProxyOptions) AddFlags(flagset *pflag.FlagSet) {
 	flagset.StringVar(&o.Upstream, "upstream", "", "The upstream URL to proxy to once requests have successfully been authenticated and authorized.")
 	flagset.BoolVar(&o.UpstreamForceH2C, "upstream-force-h2c", false, "Force h2c to communicate with the upstream. This is required when the upstream speaks h2c(http/2 cleartext - insecure variant of http/2) only. For example, go-grpc server in the insecure mode, such as helm's tiller w/o TLS, speaks h2c only")
+	flagset.IntVar(&o.UpstreamDNSTimeout, "upstream-dns-timeout", 5, "The timeout in seconds for DNS lookups of the upstream. If set to 0, no timeout is set.")
```
stlaz, line 67: People don't need to be able to configure this. Besides, it's not actually used in the described way, there's only one spot that uses it like that.

cmd/kube-rbac-proxy/app/options/proxyoptions.go
```diff
@@ -125,13 +140,87 @@ func (o *ProxyOptions) ApplyTo(c *server.KubeRBACProxyInfo, a *serverconfig.Auth
...
+	isLoopback, err := isLoopbackAddress(ctx, o.Upstream)
+	if err == nil && isLoopback {
+		return nil
+	}
+
+	if o.UpstreamForceH2C {
+		return fmt.Errorf("loopback address is required for h2c")
+	}
+
+	// isMTLSPossible emphasizes that we can't verify that upstream verifies the
```
stlaz, line 164: there's nothing like `isMTLSPossible` around :eyes:

cmd/kube-rbac-proxy/app/options/proxyoptions.go
```diff
@@ -125,13 +140,87 @@ func (o *ProxyOptions) ApplyTo(c *server.KubeRBACProxyInfo, a *serverconfig.Auth
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(o.UpstreamDNSTimeout)*time.Second)
+	defer cancel()
+
+	isLoopback, err := isLoopbackAddress(ctx, o.Upstream)
+	if err == nil && isLoopback {
+		return nil
+	}
+
+	if o.UpstreamForceH2C {
+		return fmt.Errorf("loopback address is required for h2c")
```
stlaz, line 161: the above error is unhandled, we may still be on a loopback, so this error message might be false

cmd/kube-rbac-proxy/app/options/proxyoptions.go
```diff
@@ -125,13 +140,87 @@ func (o *ProxyOptions) ApplyTo(c *server.KubeRBACProxyInfo, a *serverconfig.Auth
...
+	}
+
+	// isMTLSPossible emphasizes that we can't verify that upstream verifies the
+	// client's certificate.
+
+	if err := validateMTLSConfiguration(o.UpstreamClientCertFile, o.UpstreamClientKeyFile, o.UpstreamCAFile); err != nil {
+		return err
+	}
+
+	return nil
```
stlaz, line 171: `return validateMTLSConfiguration(o.UpstreamClientCertFile, o.UpstreamClientKeyFile, o.UpstreamCAFile)`

cmd/kube-rbac-proxy/app/options/proxyoptions.go
```diff
@@ -125,13 +140,87 @@ func (o *ProxyOptions) ApplyTo(c *server.KubeRBACProxyInfo, a *serverconfig.Auth
...
+
+func validateMTLSConfiguration(upstreamClientCertPath, upstreamClientKeyPath, upstreamCAPath string) error {
+	// Check if client is configured to provide a certificiate.
+	if len(upstreamClientCertPath) == 0 || len(upstreamClientKeyPath) == 0 || len(upstreamCAPath) == 0 {
+		return fmt.Errorf("krp client cert/key and an upstream CA are required for mTLS")
+	}
+
+	_, err := tls.LoadX509KeyPair(upstreamClientCertPath, upstreamClientKeyPath)
+	if err != nil {
+		return fmt.Errorf("failed to read upstream client cert/key: %w", err)
```
stlaz, line 182: the above does not have to be a read error, see https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/crypto/tls/tls.go;l=245

cmd/kube-rbac-proxy/app/options/proxyoptions.go
```diff
@@ -125,13 +140,87 @@ func (o *ProxyOptions) ApplyTo(c *server.KubeRBACProxyInfo, a *serverconfig.Auth
...
 
+	c.UpstreamHeaders = o.UpstreamHeader
 	c.IgnorePaths = o.IgnorePaths
 	c.AllowPaths = o.AllowPaths
 	a.APIAudiences = o.TokenAudiences
 
 	return nil
 }
 
+func validateSecureConnection(o *ProxyOptions) error {
```
stlaz, line 151: this function does not validate connections

test/kubetest/tls.go
```diff
@@ -0,0 +1,225 @@
+/*
```
stlaz, line 1: can you use any of https://github.com/openshift/kubernetes/blob/913d16a5d9fb76dfb617551d46b8c361a7128d3f/staging/src/k8s.io/client-go/util/cert/cert.go ?

test/kubetest/client.go
```diff
@@ -16,7 +16,22 @@ limitations under the License.
 
 package kubetest
 
-import "k8s.io/client-go/kubernetes"
+import (
+	"k8s.io/client-go/kubernetes"
+)
+
+func ClientLogsContain(client kubernetes.Interface, command string, logEntries []string, opts *RunOptions) Action {
```
stlaz, line 23: what does this do?

stlaz (changes requested):

Makefile
```diff
@@ -94,8 +94,12 @@ test-unit:
 test-e2e:
 	go test -timeout 55m -v ./test/e2e/ $(TEST_RUN_ARGS) --kubeconfig=$(KUBECONFIG)
 
-test-local:
-	@echo 'run: VERSION=local make clean container kind-create-cluster test'
+test-local-setup: VERSION = local
+test-local-setup: VERSION_SEMVER = $(shell cat VERSION)
+test-local-setup: clean container kind-create-cluster
```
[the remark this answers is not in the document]
ibihim: Do you want me to move the clean target out of the setup?

cmd/kube-rbac-proxy/app/options/proxyoptions.go
ibihim: Good catch.

test/kubetest/client.go
ibihim: I assume you want me to add docs above.

cmd/kube-rbac-proxy/app/options/proxyoptions.go
ibihim: I assume you want me to change the name.

I am renaming it to `validateSecureConnectionConfig`.

cmd/kube-rbac-proxy/app/options/proxyoptions.go
ibihim: Let me handle the error consciously as it might create the impression, that a mTLS connection is necessary for identity headers, even though a loopback address would be fine as well.

cmd/kube-rbac-proxy/app/options/proxyoptions.go
ibihim: If we are able to read the file, but we can't interpret the content properly, it is still a read error.

I can change the wording to be more generic.

cmd/kube-rbac-proxy/app/options/proxyoptions.go
stlaz: Since it's used only in validation, it's not really an upstream DNS timeout, as DNS lookups will be happening with all the "DNS cache miss" requests upstream, which are nowhere affected by this option.

If you feel like 5 seconds is too little, make it 10. However, if a DNS lookup is taking longer than 5s then there's something from with your infra.

cmd/kube-rbac-proxy/app/options/proxyoptions.go
stlaz: > If it is not a loopback address or there is an error, h2c shall not work and we continue validating for mTLS.
> If the config enables mTLS, we don't need loopback for identity headers.

right, but even in those cases you should at least log the error for more clarity

test/kubetest/client.go
stlaz: that, and I actually wanted to know what the function does :laughing:

Makefile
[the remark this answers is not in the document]
stlaz: I believe a cleanup might be unexpected, so yes, remove it from the flow, please.

cmd/kube-rbac-proxy/app/options/proxyoptions.go
ibihim: As it is a nit and it is not targeting anything, I see no CTA.

test/kubetest/client.go
ibihim: I configured nginx to respond with the `x-remote-user header` received. I then check the client logs, if nginx was reached with the identity header, by reading nginx's response.

cmd/kube-rbac-proxy/app/options/proxyoptions.go
```diff
@@ -125,13 +139,101 @@ func (o *ProxyOptions) ApplyTo(c *server.KubeRBACProxyInfo, a *serverconfig.Auth
...
 	return nil
 }
 
+func validateSecureConnectionConfig(o *ProxyOptions) error {
+	if !identityheaders.HasIdentityHeadersEnabled(o.UpstreamHeader) || !o.UpstreamForceH2C {
+		return nil
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(loopbackLookupTimeout)*time.Second)
+	defer cancel()
```
stlaz, line 156: this has no use here, push it inside of the `isLoopbackAddress` function

cmd/kube-rbac-proxy/app/options/proxyoptions.go
```diff
@@ -57,6 +65,7 @@ type ProxyOptions struct {
 func (o *ProxyOptions) AddFlags(flagset *pflag.FlagSet) {
 	flagset.StringVar(&o.Upstream, "upstream", "", "The upstream URL to proxy to once requests have successfully been authenticated and authorized.")
 	flagset.BoolVar(&o.UpstreamForceH2C, "upstream-force-h2c", false, "Force h2c to communicate with the upstream. This is required when the upstream speaks h2c(http/2 cleartext - insecure variant of http/2) only. For example, go-grpc server in the insecure mode, such as helm's tiller w/o TLS, speaks h2c only")
+	flagset.IntVar(&o.UpstreamDNSTimeout, "upstream-dns-timeout", 5, "The timeout in seconds for DNS lookups of the upstream. If set to 0, no timeout is set.")
```
stlaz, line 68: forgot to remove the flag from here

cmd/kube-rbac-proxy/app/options/proxyoptions.go
```diff
@@ -125,13 +139,101 @@ func (o *ProxyOptions) ApplyTo(c *server.KubeRBACProxyInfo, a *serverconfig.Auth
...
+func validateSecureConnectionConfig(o *ProxyOptions) error {
+	if !identityheaders.HasIdentityHeadersEnabled(o.UpstreamHeader) || !o.UpstreamForceH2C {
+		return nil
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(loopbackLookupTimeout)*time.Second)
+	defer cancel()
+
+	isLoopback, errLoopback := isLoopbackAddress(ctx, o.Upstream)
+	if errLoopback == nil && isLoopback {
```
stlaz, line 159: only log the error, the current error handling is super confusing

cmd/kube-rbac-proxy/app/options/proxyoptions.go
```diff
@@ -125,13 +139,101 @@ func (o *ProxyOptions) ApplyTo(c *server.KubeRBACProxyInfo, a *serverconfig.Auth
...
 
+func validateSecureConnectionConfig(o *ProxyOptions) error {
+	if !identityheaders.HasIdentityHeadersEnabled(o.UpstreamHeader) || !o.UpstreamForceH2C {
+		return nil
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(loopbackLookupTimeout)*time.Second)
+	defer cancel()
+
+	isLoopback, errLoopback := isLoopbackAddress(ctx, o.Upstream)
```
stlaz, line 158: move inside of the `if o.UpstreamForceH2C` branch?

cmd/kube-rbac-proxy/app/options/proxyoptions.go
```diff
@@ -125,13 +139,101 @@ func (o *ProxyOptions) ApplyTo(c *server.KubeRBACProxyInfo, a *serverconfig.Auth
...
+	)
+}
+
+func validateMTLSConfiguration(upstreamClientCertPath, upstreamClientKeyPath, upstreamCAPath string) error {
+	// Check if client is configured to provide a certificiate.
+	if len(upstreamClientCertPath) == 0 || len(upstreamClientKeyPath) == 0 || len(upstreamCAPath) == 0 {
+		return fmt.Errorf("krp client cert/key and an upstream CA are required for mTLS")
+	}
+
+	_, err := tls.LoadX509KeyPair(upstreamClientCertPath, upstreamClientKeyPath)
```
stlaz, line 193: aren't these validated elsewhere?

cmd/kube-rbac-proxy/app/options/proxyoptions.go
ibihim: How silly 😅

cmd/kube-rbac-proxy/app/options/proxyoptions.go
ibihim: Too late, already squashed :smile:

cmd/kube-rbac-proxy/app/options/proxyoptions.go
ibihim: This is necessary for identity headers too.

cmd/kube-rbac-proxy/app/options/proxyoptions.go
ibihim: It is being done in the Config setup. As it is only done on startup, I am accepting it, to keep the code clean.

Though process:
1. Validate is called before Complete, which makes sense. Why should you try to complete on invalid flags? So populating options with it is not feasible.
2. If Validate would fill up the Options / Config, it would violate the single purpose in its name.

WDYT?

cmd/kube-rbac-proxy/app/options/proxyoptions.go
ibihim: I improved the logic. It should be pretty clear by now.

cmd/kube-rbac-proxy/app/options/proxyoptions.go
stlaz: I don't think the current shape is correct:

You only check the correctness of the upstream CA and upstream client cert/key pair in case we've got identity headers or we're forcing H2C.
You still won't check them in case upstream is a loopback address, even though upstream may require client cert auth.

That means you've got validation that mostly doesn't work.

I see two possible ways to get out of this:
1. you always verify the options
2. you'll delegate the verification of the certs to the `ApplyTo()` code and only validate that the flag combination is correct

cmd/kube-rbac-proxy/app/options/proxyoptions.go
```diff
@@ -125,13 +138,96 @@ func (o *ProxyOptions) ApplyTo(c *server.KubeRBACProxyInfo, a *serverconfig.Auth
...
+		"loopback address or mTLS is required for identity headers: loopback - %w; mTLS - %w",
+		errLoopback,
+		errMTLS,
+	)
+}
+
+func validateMTLSConfiguration(upstreamClientCertPath, upstreamClientKeyPath, upstreamCAPath string) error {
+	// Check if client is configured to provide a certificiate.
+	if len(upstreamClientCertPath) == 0 || len(upstreamClientKeyPath) == 0 || len(upstreamCAPath) == 0 {
+		return fmt.Errorf("krp client cert/key and an upstream CA are required for mTLS")
```
stlaz, line 177: Why do you need a special CA for that?

Note that, contrary to the docs, this is not required in that case:
https://github.com/brancz/kube-rbac-proxy/blob/f5d9a22984cf08a7b16b18df8cb3343d9169e7f6/pkg/server/config.go#L84-L97

ibihim: If we have a mTLS setup, the krp client needs to have certs: check.
If we have a mTLS setup, we need a CA from the upstream server we are connecting to.

stlaz: >  If we have a mTLS setup, we need a CA from the upstream server we are connecting to.
unless it's already trusted by the system

ibihim:
```mermaidjs
sequenceDiagram
  participant krp
  participant upstream
  
  krp ->> upstream: look, this is my upstreamClientCert and I can encrypt it with upstreamClientKey.
  upstream ->> krp: wow, so cool. Btw, here is my serving certs signed by upstreamCA, you know, I am serving HTTPS.
```

cmd/kube-rbac-proxy/app/options/proxyoptions.go
```diff
@@ -125,13 +135,71 @@ func (o *ProxyOptions) ApplyTo(c *server.KubeRBACProxyInfo, a *serverconfig.Auth
...
+	}
+	if o.UpstreamForceH2C {
+		return fmt.Errorf("loopback address is required for h2c: %w", errLoopback)
+	}
+
+	// If Identity Headers are configured and it is not a loopback address,
+	// verify that mTLS is configured.
+	if len(o.UpstreamClientCertFile) == 0 || len(o.UpstreamClientKeyFile) == 0 || len(o.UpstreamCAFile) == 0 {
+		return fmt.Errorf(
+			"loopback address or mTLS is required for identity headers: loopback - %w; mTLS - krp client cert/key and an upstream CA are required for mTLS",
```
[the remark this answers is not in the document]
ibihim: I like my message more 😆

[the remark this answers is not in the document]
stlaz: Me, a common user: "what's a krp?"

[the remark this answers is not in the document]
stlaz: Also, why do you show the error from checking the localhost url rather than the URL that the user configured?

cmd/kube-rbac-proxy/app/options/proxyoptions.go
```diff
@@ -125,13 +135,76 @@ func (o *ProxyOptions) ApplyTo(c *server.KubeRBACProxyInfo, a *serverconfig.Auth
...
 
+func validateSecureConnectionConfig(o *ProxyOptions) error {
+	if !identityheaders.HasIdentityHeadersEnabled(o.UpstreamHeader) && !o.UpstreamForceH2C {
+		return nil
+	}
+
+	errLoopback := validateLoopbackAddress(o.Upstream)
+	if errLoopback == nil {
+		return nil
+	}
```
stlaz, line 154: debug log the error here instead of including it in the err validation msg later

ibihim: Currently you get one message that tells you the issue, rather than finding the mTLS err and then searching if there was a loopback err.

Isn't this better to debug?

cmd/kube-rbac-proxy/app/options/proxyoptions.go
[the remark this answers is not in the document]
ibihim: ok, I didn't see that you provide `o.Upstream`.

Makefile
```diff
@@ -96,8 +96,8 @@ test-e2e:
 
 test-local-setup: VERSION = local
 test-local-setup: VERSION_SEMVER = $(shell cat VERSION)
-test-local-setup: clean container kind-create-cluster
-test-local: test-local-setup test-e2e
+test-local-setup: container kind-create-cluster
```
stlaz, line 99: wrong commit

stlaz (approved):

PR #285, opened 2024-03-18
add command line args --kube-api-qps/--kube-api-burst

novegit:
Using kube-rbac-proxy in front of a prometheus in a multi tenant environment causes often client-side throttling events, that slow down connections, when more than 10 queries per second  are coming in. Only way to mitigate is to increase the number of replicas of the kube-rbac-proxy pods.

**Example for client-side throttling**:
I0318 12:32:18.188838     800 request.go:629] Waited for 993.646183ms due to client-side throttling, not priority and fairness, request: POST:https://172.30.0.1:443/apis/authorization.k8s.io/v1/subjectaccessreviews
I0318 12:32:18.192938     800 round_trippers.go:553] POST https://172.30.0.1:443/apis/authorization.k8s.io/v1/subjectaccessreviews 201 Created in 3 milliseconds
I0318 12:32:19.189607     800 request.go:629] Waited for 1.994365552s due to client-side throttling, not priority and fairness, request: POST:https://172.30.0.1:443/apis/authorization.k8s.io/v1/subjectaccessreviews

with the new parameters, the queries per second can be set above the default kube-client default value of 10 qps.
--kube-api-qps
--kube-api-burst

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,8 @@ Kube-rbac-proxy flags:
       --http2-max-size uint32                       The maximum number of bytes that the server will accept for frame size and buffer per stream in a HTTP/2 request. (default 262144)
       --ignore-paths strings                        Comma-separated list of paths against which kube-rbac-proxy pattern-matches the incoming request. If the requst matches, it will proxy the request without performing an authentication or authorization check. Cannot be used with --allow-paths.
       --insecure-listen-address string              [DEPRECATED] The address the kube-rbac-proxy HTTP server should listen on.
+      --kube-api-burst int                          kube-api burst value; needed when kube-api-qps is set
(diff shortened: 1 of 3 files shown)


ibihim:
@novegit, those are the same flags as used in kube, right?

Cool.

ibihim:
@novegit, nice try 😆  You need to run `make generate` and commit the README.md changes.

novegit:
should work now.

novegit:
> @novegit, those are the same flags as used in kube, right?
> 
> Cool. Thx for the contribution.

yes, on my first try, I only added 'qps'. An an error message appeared, that also 'burst' must be set, when 'qps' is set.

PR #287, opened 2024-03-25
Bump krp to k8s 1.30

ibihim:
## What

Bump dependencies to higher versions. Most prominent: k8s v1.29.3

## Why

It makes sense to stay up to date with deps.
In particular it should fix the CVE reports about telemetry.io


pkg/authn/oidc.go
```diff
@@ -43,13 +44,23 @@ func NewOIDCAuthenticator(config *OIDCConfig) (*OIDCAuthenticator, error) {
...
+				Username: apiserver.PrefixedClaimOrExpression{
+					Prefix: &config.UsernamePrefix,
+					Claim:  config.UsernameClaim,
+				},
+				Groups: apiserver.PrefixedClaimOrExpression{
+					Prefix: &config.GroupsPrefix,
+					Claim:  config.GroupsClaim,
+				},
+			},
+		},
```
aramase, line 62: This change looks good!

aramase (commented):
Just one comment about the `golang.org/x/net` version

README.md
```diff
@@ -77,6 +77,7 @@ Kube-rbac-proxy flags:
       --oidc-issuer string                          The URL of the OpenID issuer, only HTTPS scheme will be accepted. If set, it will be used to verify the OIDC JSON Web Token (JWT).
       --oidc-sign-alg stringArray                   Supported signing algorithms, default RS256 (default [RS256])
       --oidc-username-claim string                  Identifier of the user in JWT claim, by default set to 'email' (default "email")
+      --oidc-username-prefix string                 If provided, the username will be prefixed with this value to prevent conflicts with other authentication strategies.
```
stlaz, line 80: do we actually need to add the prefix?

cmd/kube-rbac-proxy/app/options/options.go
```diff
@@ -120,6 +120,7 @@ func (o *ProxyRunOptions) Flags() k8sapiflag.NamedFlagSets {
 	flagset.StringVar(&o.Auth.Authentication.OIDC.ClientID, "oidc-clientID", "", "The client ID for the OpenID Connect client, must be set if oidc-issuer-url is set.")
 	flagset.StringVar(&o.Auth.Authentication.OIDC.GroupsClaim, "oidc-groups-claim", "groups", "Identifier of groups in JWT claim, by default set to 'groups'")
 	flagset.StringVar(&o.Auth.Authentication.OIDC.UsernameClaim, "oidc-username-claim", "email", "Identifier of the user in JWT claim, by default set to 'email'")
+	flagset.StringVar(&o.Auth.Authentication.OIDC.UsernamePrefix, "oidc-username-prefix", "", "If provided, the username will be prefixed with this value to prevent conflicts with other authentication strategies.")
 	flagset.StringVar(&o.Auth.Authentication.OIDC.GroupsPrefix, "oidc-groups-prefix", "", "If provided, all groups will be prefixed with this value to prevent conflicts with other authentication strategies.")
 	flagset.StringArrayVar(&o.Auth.Authentication.OIDC.SupportedSigningAlgs, "oidc-sign-alg", []string{"RS256"}, "Supported signing algorithms, default RS256")
```
stlaz, line 125: should we mention deprecation of this flag as in upstream?

pkg/authn/oidc.go
```diff
@@ -36,20 +37,30 @@ var (
...
 	dyCA, err := dynamiccertificates.NewDynamicCAContentFromFile("oidc-ca", config.CAFile)
 	if err != nil {
 		return nil, err
 	}
 
-	tokenAuthenticator, err := oidc.New(oidc.Options{
-		IssuerURL:            config.IssuerURL,
-		ClientID:             config.ClientID,
+	tokenAuthenticator, err := oidc.New(ctx, oidc.Options{
+		JWTAuthenticator: apiserver.JWTAuthenticator{
```
stlaz, line 47: do we get the dynamic CA for free when we move it here instead of using it as below?

README.md
ibihim: I assume that we don't need it as no one ever complained. It felt like a natural thing to do as we have `--oidc-groups-prefix`.

cmd/kube-rbac-proxy/app/options/options.go
ibihim: Yes, we should. Good point.

ibihim: Where is the marker for deprecation?

pkg/authn/oidc.go
ibihim: Which of the properties enables a dynamic CA from within JWTAuthenticator?

ibihim: The dynamic CA content provider is being used to create the client.

https://github.com/kubernetes/kubernetes/blob/9d5db28f5f5d5a35012dd6162af374420e6e1f4c/staging/src/k8s.io/apiserver/plugin/pkg/authenticator/token/oidc/oidc.go#L270-L291

cmd/kube-rbac-proxy/app/options/options.go
stlaz: Interesting. Looks like the only place where this is mentioned is this (recent) enhancement: https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/3331-structured-authentication-configuration#goals

Feel free to ignore my original comment, then.

njyeti:
@ibihim when are you planning for the next release with these cves fixes?

PR #288, opened 2024-03-28
Add dependabot

lucacome:
Adds dependabot to keep all the dependencies up to date


lucacome:
@ibihim what do you think about this?

stlaz:
I would rather avoid spam PRs for dependencies. We'll stick to the security warnings we're already getting for the repo.

PR #293, opened 2024-05-07
 fix bool,deprecated flags regression

ibihim:
## What

Fix a regression for the removed flags.

## Why

If the bool value flag was specified like `--logtostderr`, without the `=true`, it was treated like `--logtostderr=true` in v0.15.0.
In v0.16.0 and v0.17.0, `--logtostderr` without any value would treat the follow up flag as its value, which leads to a regression.


liouk (changes requested):
I would split the changes of `cmd/kube-rbac-proxy/app/kube-rbac-proxy.go` to a separate commit as this fixes a different issue.

PR #296, opened 2024-05-07
Cut release v0.16.1

ibihim:
## What

Create a patch for 0.16.0

## Why

There was a regression in the flag behavior.


ibihim:
Will get pushed as v0.16.1 branch

Issue #196, opened 2022-08-26
[Deprecation] Removal of options and features that don't fit with upstream


ibihim:
# What

We will remove the following options in the v1 release:
1. `--tls-reload-interval` as this option is not supported by upstream,
2. `insecure-listen-addresss` won't work anymore and
3. `tls-cert-file == nil` and `tls-private-key-file == nil` won't work any more, TLS is now mandatory.

# Why

## 1.

We were asked to use the upstream cert loader, which doesn't support custom intervals.

## 2. and 3.

We are aware that we create obstacles in running kube-rbac-proxy for testing or debugging purposes.
But we reduce the probability for an insecure set up of kube-rbac-proxy, which is a security relevant component.

Running kube-rbac-proxy without TLS certificates, makes it possible to impersonate kube-rbac-proxy.

The reason that we remove that capability is a pre-acceptance requirement for kube-rbac-proxy, before we can donate the project so sig-auth of k8s.

# Reference

- https://github.com/brancz/kube-rbac-proxy/issues/169
- https://github.com/brancz/kube-rbac-proxy/pull/197
- https://github.com/brancz/kube-rbac-proxy/issues/169
- https://github.com/brancz/kube-rbac-proxy/pull/188
- https://github.com/brancz/kube-rbac-proxy/pull/186
- https://github.com/brancz/kube-rbac-proxy/issues/187

camilamacedo86:
HI @ibihim,

Could you please add here how we should replace those flags to sort out the deprecation?

See that we are using:

```ymal
- "--secure-listen-address=0.0.0.0:8443"
        - "--upstream=http://127.0.0.1:8080/"
        - "--logtostderr=true"
        - "--v=0" 
```

So, how it should be done now?

ibihim:
I will add a README.md entry for this.

PR #299, opened 2024-06-17
doc: Note that tls-reload-interval is deprecated

hoexter:
Seems this was depcreated as part of
https://github.com/brancz/kube-rbac-proxy/issues/196 and a log message warns about it.


am3o (approved):

am3o:
@ibihim  Can you check it out?

stlaz:
lgtm

PR #300, opened 2024-06-28
Introduce changes from master to sig-auth-acceptance branch

ibihim:
## What

- Version bump sig-auth-acceptance branch.
- Introduce important changes from master to sig-auth-acceptance branch.
- Make e2e-tests work on arm64 machines.

## Why

- For some of the change request a newer version of k8s is required, like the wrapper around token audiences.
- The http/2 disablement is important from master for one of the change requests.
- Reduce differences in the Makefile to have a smaller context change.
- Enable e2e tests on ARM64, in particular for MacOS M-Chip users.


.github/workflows/build.yml
```diff
@@ -4,8 +4,8 @@ on: [push, pull_request]
 
 env:
   QUAY_PATH: quay.io/brancz/kube-rbac-proxy
-  go-version: '1.22.1'
-  kind-version: 'v0.22.0'
+  go-version: '1.22.4'
+  kind-version: 'v0.20.0'
```
liouk, line 8: `kind` downgrade intentional?

README.md
```diff
@@ -97,6 +97,7 @@ Proxy flags:
       --auth-header-user-field-name string          The name of the field inside a http(2) request header to tell the upstream server about the user's name (default "x-remote-user")
       --auth-token-audiences strings                Comma-separated list of token audiences to accept. By default a token does not have to have any specific audience. It is recommended to set a specific audience.
       --config-file string                          Configuration file to configure static and rewrites authorization of the kube-rbac-proxy.
+      --disable-http2-serving                       If true, HTTP2 serving will be disabled [default=false]
```
liouk, line 100: This should be part of commit 48d0d9cfa185f8e4831a588b5a43c6b921585061

liouk (changes requested):

.github/workflows/build.yml
ibihim: no

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
```diff
@@ -232,7 +232,7 @@ func Run(cfg *server.KubeRBACProxyConfig) error {
 
 	handler := identityheaders.WithAuthHeaders(proxy, cfg.KubeRBACProxyInfo.UpstreamHeaders)
 	handler = kubefilters.WithAuthorization(handler, authz, scheme.Codecs)
-	handler = kubefilters.WithAuthentication(handler, authenticator, http.HandlerFunc(filters.UnauthorizedHandler), cfg.DelegatingAuthentication.APIAudiences)
+	handler = kubefilters.WithAuthentication(handler, authenticator, http.HandlerFunc(filters.UnauthorizedHandler), cfg.DelegatingAuthentication.APIAudiences, nil)
```
stlaz, line 235: What happens when you supply the request header nil here? It won't get autodetected from the cluster, right?

pkg/authn/oidc.go
```diff
@@ -36,20 +37,30 @@ var (
...
 	dyCA, err := dynamiccertificates.NewDynamicCAContentFromFile("oidc-ca", config.CAFile)
 	if err != nil {
 		return nil, err
 	}
 
-	tokenAuthenticator, err := oidc.New(oidc.Options{
-		IssuerURL:            config.IssuerURL,
-		ClientID:             config.ClientID,
+	tokenAuthenticator, err := oidc.New(ctx, oidc.Options{
+		JWTAuthenticator: apiserver.JWTAuthenticator{
```
stlaz, line 47: can we remove the old options and just wire the new config, since we have the chance?

cmd/kube-rbac-proxy/app/options/proxyoptions.go
```diff
@@ -116,29 +120,30 @@ func (o *ProxyOptions) Validate() []error {
 	return errs
 }
 
-func (o *ProxyOptions) ApplyTo(c *server.KubeRBACProxyInfo, a *serverconfig.AuthenticationInfo) error {
+func (o *ProxyOptions) ApplyTo(config *server.KubeRBACProxyConfig) error {
```
stlaz, line 123: The previous was better in terms of isolation.

stlaz (commented):
much smaller change than I thought it would be :+1:

cmd/kube-rbac-proxy/app/options/proxyoptions.go
ibihim: Yes, I know, but I need it to set up HTTP/2 disablement. We can revert it in k8s 1.31.

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
ibihim: TODO add explanation of what nil does.

ibihim: It enables you to cleanse custom headers that you use for your proxy to contain additional information. It is necessary to drop those.

We write custom headers in case of request headers and we read custom headers based on the rewrite logic: [link](https://github.com/ibihim/kube-rbac-proxy/blob/f31c1870b1800a92f0f37c01b7a20c17adfc5412/pkg/authorization/rewrite/middleware.go#L56-L63).
The rewrite logic for Query parameters relies on upstream to interpret them, which is usually a hostile behavior that you want to avoid, but in our case, we want Prometheus to be able to interpret it. Therefore I am hesitant to introduce it for custom headers as well. We could add a boolean flag that defaults to true, which enables / disables cleansing of header values? **WDYT @stlaz?**

Ref: [k8s.io/apiserver/pkg/endpoints/filters/authentication.go](https://github.com/kubernetes/kubernetes/blob/1904a56e3071035de19016cef621866a37140d5f/staging/src/k8s.io/apiserver/pkg/endpoints/filters/authentication.go#L98-L105)

pkg/authn/oidc.go
ibihim: @stlaz, WDYT?

https://github.com/brancz/kube-rbac-proxy/pull/300/commits/a610979bafbaa50c79e334ebe3dc92aa7e77ca2f

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
stlaz: We should avoid forwarding the default headers at least, which is what `WithAuthentication()`does already.

Generally I think we should drop any headers we're processing, more so if we're using them for authentication. I don't think we wired any authn via request headers, correct?

Outside of that we should also make sure that none of the headers we produce can be just passed upstream from the original request.

pkg/authn/oidc.go
ibihim: I understand. Let me remove the change.

Introducing the suggested change, is definitely another PR.

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
ibihim: Let me summarize:

- `WithAuthentication` does a great job.
- We should remove headers we consume, which would include the rewrites rule in the HEADER. This task would include some research on the impact on Prometheus.
- If identityheaders are enabled, we should remove those headers if set by the client. This is also a request from the review by Mo.

This would be another PR.

cmd/kube-rbac-proxy/app/options/proxyoptions.go
ibihim: No, we set more than the secure serving, we also have the krp info and delegated authn that contains authn / authz and routing information..

stlaz: I meant on top of all the other things that were being passed in previously

ibihim: Yea, but like 3 arguments? Why not handing in the whole object then?

stlaz: to isolate the changes, and also to stress that the function probably does more things that it should and we may want to refactor it in the near future

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
stlaz: At least add a FIXME here then, I think we need to address this before this makes it to production

pkg/authn/oidc.go
stlaz: would you create a k/k issue and link it to this PR, please?

cmd/kube-rbac-proxy/app/options/proxyoptions.go
ibihim: Yes yes, lets keep it ugly, such that feel the itch to make it better soon.

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
ibihim: Fixing before production? 100%. I can add a FIXME.

pkg/authn/identityheaders/identityheaders_test.go
```diff
@@ -46,13 +46,27 @@ func TestWithAuthHeaders(t *testing.T) {
...
 		header map[string][]string
 	}{
 		{
-			name:   "should pass through",
-			cfg:    &identityheaders.AuthnHeaderConfig{},
-			header: map[string][]string{},
+			name: "should pass through",
+			cfg:  &identityheaders.AuthnHeaderConfig{},
+			req: func() *http.Request {
+				req, err := http.NewRequest(http.MethodGet, "http://example.com", nil)
```
stlaz, line 56: Add a set of chaining functions that construct a request as necessary

pkg/authn/identityheaders/identityheaders.go
```diff
@@ -43,6 +43,10 @@ func WithAuthHeaders(handler http.Handler, cfg *AuthnHeaderConfig) http.Handler
 	}
 
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		// We don't want the client to impersonate someone else.
+		req.Header.Del(cfg.UserFieldName)
```
stlaz, line 47: this was nearly a CVE ->backport to master

ibihim: It was part of the review I had with @enj. It wouldn't have been forgotten.

pkg/authn/oidc.go
ibihim: What do you expect me to create an issue for? That we finally need to move that into one place instead of copying it across different places and that this PR is an example of why we need it?

cmd/kube-rbac-proxy/app/options/proxyoptions.go
ibihim: Sorry, I must have dropped it with git.

pkg/authn/identityheaders/identityheaders_test.go
stlaz: removal of duplicated code + it's easier to write future unit tests

stlaz:
please squash the last two changes to where the code was originally introduced and let's merge this PR

PR #307, opened 2024-08-29
Implement method not allowed response

jpinsonneau:
Currently, auth return a `403` error whenever the user have other access or not.

This PR suggest to add `405` status code when authorization config contains another verb than the one needed.

This will allow usage of golang prometheus client that actually send a `POST` query first before retrying on a `GET` one if the response code is `405 = StatusMethodNotAllowed`
https://github.com/prometheus/client_golang/blob/main/api/prometheus/v1/api.go#L1436-L1462

### DIFF
--- a/pkg/authz/auth.go
+++ b/pkg/authz/auth.go
@@ -104,7 +104,7 @@ type staticAuthorizer struct {
 	config []StaticAuthorizationConfig
 }
 
-func (saConfig StaticAuthorizationConfig) Matches(a authorizer.Attributes) bool {
+func (saConfig StaticAuthorizationConfig) Matches(a authorizer.Attributes) (bool, bool) {
 	isAllowed := func(staticConf string, requestVal string) bool {
 		if staticConf == "" {
 			return true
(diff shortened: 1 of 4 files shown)


stlaz:
The job of a Kubernetes authorizer is to say whether a request was allowed or whether to delegate authorization further. There should be no "if you try your request a little differently, maybe I'll let you in".

Why would a client not know how to reach an endpoint?

stlaz:
> It seems to be a convention that servers answer 405 error in that specific case.

The server answers 405 in case it's not handling such methods at the given endpoint. If the implementation is kind to the client, it would even let you know which methods to try. However, the kube-rbac-proxy does handle it. Later, the internal logic transfers it into Kubernetes context where it pops out as a `verb` for the internal authorization system. 

The 403 response in this case is a response to the user _not being allowed_ to perform an action but does not mean the given method is not being served at the endpoint.

You can either fix the prometheus client to behave reasonably so that it doesn't try to guess how to query an endpoint, or allow the verbs you need in your static authorizer config.

stlaz:
I am going to close this PR now, assuming the explanation helped.

PR #308, opened 2024-08-30
Makefile: fix arm64 on MacOS

ibihim:
## What

Fixate Linux as OS.

## Why

No matter what the CPU architecture is, the tests will run within a container that is Linux. On MacOS there is a Linux VM being run in the background that uses the container runtimes.


ibihim:
duplicate

PR #311, opened 2024-09-11
Mask tokens for logging in kube-rbac-proxy

kramaranya:


cmd/kube-rbac-proxy/app/sanitazion.go
```diff
@@ -0,0 +1,70 @@
...
+			}
+		}
+	}
+	return args
+}
+
+// FilterF is the filter function for formatting logging functions of klog.
+func (sf *SanitizingFilter) FilterF(format string, args []interface{}) (string, []interface{}) {
+	for i, v := range args {
+		if strValue, ok := v.(string); ok {
```
ibihim, line 26: Golang encourages us to view at this as `[]byte`. How would the code change, if we would:

```suggestion
		if strValue, ok := v.([]byte); ok {
```

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
```diff
@@ -218,6 +218,8 @@ func Run(cfg *completedProxyRunOptions) error {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
+	klog.SetLogFilter(&SanitizingFilter{})
```
ibihim, line 221: I wouldn't put it into the `Run` function. IRather in the `NewKubeRBACProxyCommand`. Maybe in `PersistentPreRunE`.

ibihim (commented):
Could you add e2e-tests ideally or at least unit tests?

cmd/kube-rbac-proxy/app/sanitazion.go
kramaranya: @ibihim I agree with you, but in this case all of our logs are stored and processed as `string`. converting them to `[]byte` will add complexity as we'd have to convert them back to `string` for log output anyway. wdyt?

ibihim: Sure, then lets change the other functions.

cmd/kube-rbac-proxy/app/sanitazion.go
```diff
@@ -0,0 +1,70 @@
...
+			if containsTokenReview(strValue) {
+				keysAndValues[i] = maskTokenInLog(strValue)
+			}
+		}
+	}
+	return msg, keysAndValues
+}
+
+func containsTokenReview(logStr string) bool {
+	return bytes.Contains([]byte(logStr), []byte(`"kind":"TokenReview"`))
```
ibihim, line 48: Why not `strings.Contains`?

cmd/kube-rbac-proxy/app/sanitazion.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func maskTokenInLog(logStr string) string {
+	var logMap map[string]interface{}
+	if err := json.Unmarshal([]byte(logStr), &logMap); err != nil {
+		return logStr
```
ibihim, line 54: So `containsTokenReview` is true, which means that there is most probably a token and if we fail to unmarshal it, we return the unmasked string. Wouldn't it be better to just return "<failed masking>" and sacrifice the whole content, rather than leaking a token?

cmd/kube-rbac-proxy/app/sanitazion.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	if spec, ok := logMap["spec"].(map[string]interface{}); ok {
```
ibihim, line 57: This is pretty static. It will always try to go for `.spec.token` and try to mask it. So in order for this to be a valid assumption, we would need to go through the k8s.io API and verify that all tokens that can be delivered over the wire, are in this path... otherwise a regex might be better. The regex is safer, but the static path is most probably quicker.

kramaranya: I’ve reviewed all the api-resources, and it turns out that only `TokenReview` contains a token field, so we can keep the current approach

cmd/kube-rbac-proxy/app/sanitazion.go
```diff
@@ -0,0 +1,70 @@
...
+				keysAndValues[i] = maskTokenInLog(strValue)
+			}
+		}
+	}
+	return msg, keysAndValues
+}
+
+func containsTokenReview(logStr string) bool {
+	return strings.Contains(logStr, `"kind":"TokenReview"`)
+}
```
ibihim, line 51: Nit: this is quite shallow. It would make more sense to the code as is into the if clause.

(nit isn't a change required)

kramaranya:
> Could you add e2e-tests ideally or at least unit tests?

@ibihim I've added an e2e test

cmd/kube-rbac-proxy/app/sanitazion.go
```diff
@@ -0,0 +1,66 @@
+package app
```
ibihim, line 1: Add a license here 😄

test/e2e/tokenmasking.go
```diff
@@ -0,0 +1,113 @@
+package e2e
```
ibihim, line 1: I don't know if we need a license here too, but could be.

test/e2e/tokenmasking.go
```diff
@@ -0,0 +1,113 @@
...
+					command,
+					nil,
+				),
+				checkLogsForMaskedToken(client),
+			),
+		}.Run(t)
+	}
+}
+
+func checkLogsForMaskedToken(client kubernetes.Interface) kubetest.Action {
```
ibihim, line 61: Man, this is great. This looks better than what I've done on the other branch 😄

test/e2e/tokenmasking.go
```diff
@@ -0,0 +1,113 @@
...
+			return fmt.Errorf("failed to get logs: %v", err)
+		}
+		defer func() {
+			closeErr := logsStream.Close()
+			if closeErr != nil && err == nil {
+				err = fmt.Errorf("failed to close logs stream: %v", closeErr)
+			}
+		}()
+
+		scanner := bufio.NewScanner(logsStream)
```
ibihim, line 90: Everything is a stream 👍

cmd/kube-rbac-proxy/app/sanitazion.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+// SanitizingFilter implements the LogFilter interface from klog with custom functions to detect and mask tokens.
+type SanitizingFilter struct{}
+
+// Filter is the filter function for non-formatting logging functions of klog.
+func (sf *SanitizingFilter) Filter(args []interface{}) []interface{} {
+	for i, v := range args {
+		if strValue, ok := v.(string); ok {
+			if strings.Contains(strValue, `"kind":"TokenReview"`) {
```
ibihim, line 17: I am curious if there are certain optimizations for this possible:
- Enable this Filter only when logging is set to a certain verbosity, which cause the tokens to be logged in the first place.
- Check for prefix `{"apiVersion": "authentication.k8s.io/v1", "kind: TokenReview"` instead of `"kind":"TokenReview"` in the whole text.
- And if we could use the json encoding error, would be more performant / sufficient (then we would pass the TokenReview object into maskTokenInLog).

kramaranya: Thanks for your suggestions! I thought about each one:

- I believe it's important to always have the filter active to ensure tokens are masked in logs all time, so that we prevent any accidnetal token exposure, even if the logging level changes.
- If I'm not mistaken logs might not always start with the same prefix or have fields in the same order.
- If I understood correctly what you mean then I think it'll slow down the performance (especially if most lines aren't `TokenReview` logs). When we check for the substring first-- we skip irrelevant logs and only process the ones that matter.

I would keep the current solution, but let me know what you think :)

ibihim: Sure lets go ahead with the current solution.

Issue #320, opened 2024-11-12
(14) Validate if a simple response by the health endpoint is enough to proof healthiness link


ibihim:
## What

It might make sense to improve the health endpoint with something more than being able to handle requests. Alive vs Ready.

[link](https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-8796db11a6b137095413f7a7109dd01393e63d425454fc04647b98d2b8c0115cR232)

Issue #244, opened 2023-06-21
Add livenessProbe for kube-rbac-proxy


jessehu:
When using kube-rbac-proxy v0.14.1, sometimes we found kube-rbac-proxy container stuck in `TLS handshake error` and can not recover automatically until it's restarted manually.
```
I0621 17:39:33.819787    1876 log.go:198] http: TLS handshake error from 10.255.9.20:55542: write tcp 10.255.9.20:9100->10.255.9.20:55542: write: broken pipe
I0621 17:40:40.388479    1876 log.go:198] http: TLS handshake error from 10.255.9.26:1531: write tcp 10.255.9.20:9100->10.255.9.26:1531: write: broken pipe
I0621 17:44:32.288256    1876 log.go:198] http: TLS handshake error from 10.255.9.26:30302: write tcp 10.255.9.20:9100->10.255.9.26:30302: write: broken pipe
```
Adding a livenessProbe can automatically restart kube-rbac-proxy container, e.g.:
```
        - image: myregistry.io/kube-rbac-proxy:v0.14.1-with-curl
          livenessProbe:
            exec:
              command:
                - /bin/sh
                - -c
                - "curl -sSL -ik -H \"Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)\" https://${IP}:9100/metrics | grep -e 'HTTP/2 200'"
            initialDelaySeconds: 30
            failureThreshold: 3
            periodSeconds: 30
            successThreshold: 1
            timeoutSeconds: 3
```

peterbueschel:
Hi, 

I would also vote for either such an option to probe the proxy (but that implies an extra binary in the image) or to maybe panic on such an error.

In our case we also receive:

```
webhook.go:154] Failed to make webhook authenticator request: Post ".....:443/apis/authentication.k8s.io/v1/tokenreviews": context deadline exceeded 
```

Afterwards only a manual restart helps to fix that...

Waji-97:
I have got the same issue using kube-rbac-proxy container with node-exporter in a single pod. To be honest, I am unsure why it occurs but whenever it does, it uses almost all of the CPU & Mem resources limited to the kube-rbac-proxy container and goes back to normal once the container is restarted. Is there any new news on this particular issue?

ibihim:
https://github.com/brancz/kube-rbac-proxy/issues/320

huangcuiyang:
[...]

PR #321, opened 2024-11-13
improve proxy healthz checking

stlaz:
This adds and internal conn check between the proxy-endpoints and main-proxy listener.

Fixes https://github.com/brancz/kube-rbac-proxy/issues/320

### DIFF
--- a/cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
+++ b/cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
@@ -249,6 +249,7 @@ func Run(cfg *server.KubeRBACProxyConfig) error {
 		return err
 	}
 
+	var stoppedCh, listenerStoppedCh <-chan struct{}
 	go func() {
 		defer wg.Done()
 		defer cancel()
@@ -263,11 +264,15 @@ func Run(cfg *server.KubeRBACProxyConfig) error {
 		<-stoppedCh
 	}()
 
+	cfg.SecureServing.Cert.CurrentCertKeyContent()
+
 	if cfg.KubeRBACProxyInfo.ProxyEndpointsSecureServing != nil {
 		// we need a second listener in order to serve proxy-specific endpoints
 		// on a different port (--proxy-endpoints-port)
 		proxyEndpointsMux := http.NewServeMux()
-		proxyEndpointsMux.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) { _, _ = w.Write([]byte("ok")) })
+		proxyEndpointsMux.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
+			proxyHealtzCheck(w, cfg.SecureServing.Listener.Addr().String(), listenerStoppedCh, stoppedCh)
+		})
 
 		if err := wg.Add(1); err != nil {
 			return err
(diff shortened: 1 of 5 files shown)


cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
```diff
@@ -263,11 +264,15 @@ func Run(cfg *server.KubeRBACProxyConfig) error {
...
 	}()
 
+	cfg.SecureServing.Cert.CurrentCertKeyContent()
+
 	if cfg.KubeRBACProxyInfo.ProxyEndpointsSecureServing != nil {
 		// we need a second listener in order to serve proxy-specific endpoints
 		// on a different port (--proxy-endpoints-port)
 		proxyEndpointsMux := http.NewServeMux()
-		proxyEndpointsMux.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) { _, _ = w.Write([]byte("ok")) })
+		proxyEndpointsMux.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
```
ibihim, line 273: I think the previous `/healthz` was completely fine. When the proxy go routine would finish, it would also cancel the health endpoint.

What you added might be worth to be used as `liveness` or whatever you call it. But the question is, if this makes sense or if this is just testing go std lib. Maybe we could ask @enj, if this makes sense.

enj: I am unsure of exactly what this trying to do, but I do think that you need something better than just "always return ok."  For example, I would at a minimum expect a check that asserts that the upstream is working.  I don't know if checking this specific listener is the correct approach though.  From the YAML below I think this is acting as a readiness check and not a health check?

test/e2e/basics.go
```diff
@@ -519,3 +519,47 @@ func testIgnorePaths(client kubernetes.Interface) kubetest.TestSuite {
 		}.Run(t)
 	}
 }
+
+func testHealthz(client kubernetes.Interface) kubetest.TestSuite {
```
enj, line 523: A test for the failure case would make it more clear what this is checking.

enj (commented):
First pass.

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
stlaz: I was trying to set up a few checks that make sure that the other listener is capable of responding to requests.

I am not sure we want to add an upstream healthz check here - the proxy info port is supposed to serve information about the proxy itself. The healthz of upstream should probably be checked directly at upstream (or should be accessed via a permissive proxied path).

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
```diff
@@ -354,3 +359,31 @@ func setupAuthorizer(krbInfo *server.KubeRBACProxyInfo, delegatedAuthz *serverco
...
+		return
+	case <-listenerStoppedChan:
+		http.Error(w, "listener stopped", http.StatusServiceUnavailable)
+		return
+	default:
+	}
+
+	// we need the tls.Dialer otherwise the server would log EOF for TLS handshakes
+	// since the connection would be cut before that was ever attempted
+	dialer := tls.Dialer{NetDialer: &net.Dialer{}, Config: &tls.Config{InsecureSkipVerify: true}}
```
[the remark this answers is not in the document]
stlaz: This is a loopback check, the serving cert's not likely to contain information for that endpoint. The check is more or less an HTTPS ping.
Maybe we could use SNI if we think we need to verify the certs, too.

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
ibihim: The previous health check is fine. This would be more like a `readiness` check?

At the beginning I had the opinion that @enj had, that we would need to "somehow check upstream", but your arguments are also right: this is the concern of upstream itself.
Now, I have no strong opinions. I see the reason behind a connection check and I am still not sure if adds enough usefulness.

ibihim:
https://github.com/brancz/kube-rbac-proxy/issues/244

Issue #319, opened 2024-11-12
(8) Verify that custom Connection headers can't drop headers link.


ibihim:
## What

Verify that custom connection header can't drop important headers. Not completely sure any more if it refers to bad config or the client trying to bypass something.

[link](https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-8796db11a6b137095413f7a7109dd01393e63d425454fc04647b98d2b8c0115cR194)

PR #324, opened 2024-11-14
reverse proxy: use Rewrite instead of Director

stlaz:
This is to mitigate Golang's behavior with `ReverseProxy.Director` when, after the outgoing request is mutated, any headers specified by "Connection" would get rmoved from the mutated request. `ReverseProxy.Rewrite()` is only called once the headers specified in "Connection" were already removed from the outgoing-to-be request.

Fixes https://github.com/brancz/kube-rbac-proxy/issues/319

*edit*: useful resources for more context: 
- https://www.ory.sh/hop-by-hop-header-vulnerability-go-standard-library-reverse-proxy/
- https://cs.opensource.google/go/go/+/refs/tags/go1.23.3:src/net/http/httputil/reverseproxy.go;l=386-435

### DIFF
--- a/cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
+++ b/cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
@@ -212,23 +212,7 @@ func Run(cfg *server.KubeRBACProxyConfig) error {
 		return fmt.Errorf("failed to setup an authorizer: %v", err)
 	}
 
-	proxy := httputil.NewSingleHostReverseProxy(cfg.KubeRBACProxyInfo.UpstreamURL)
-	proxy.Transport = cfg.KubeRBACProxyInfo.UpstreamTransport
-
-	if cfg.KubeRBACProxyInfo.UpstreamForceH2C {
-		// Force http/2 for connections to the upstream i.e. do not start with HTTP1.1 UPGRADE req to
-		// initialize http/2 session.
-		// See https://github.com/golang/go/issues/14141#issuecomment-219212895 for more context
-		proxy.Transport = &http2.Transport{
-			// Allow http schema. This doesn't automatically disable TLS
-			AllowHTTP: true,
-			// Do disable TLS.
-			// In combination with the schema check above. We could enforce h2c against the upstream server
-			DialTLS: func(netw, addr string, cfg *tls.Config) (net.Conn, error) {
-				return net.Dial(netw, addr)
-			},
-		}
-	}
+	proxy := setupProxyHandler(cfg.KubeRBACProxyInfo)
 
 	handler := identityheaders.WithAuthHeaders(proxy, cfg.KubeRBACProxyInfo.UpstreamHeaders)
 	handler = kubefilters.WithAuthorization(handler, authz, scheme.Codecs)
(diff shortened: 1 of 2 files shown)


cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
```diff
@@ -212,7 +212,15 @@ func Run(cfg *server.KubeRBACProxyConfig) error {
 		return fmt.Errorf("failed to setup an authorizer: %v", err)
 	}
 
-	proxy := httputil.NewSingleHostReverseProxy(cfg.KubeRBACProxyInfo.UpstreamURL)
+	proxy := &httputil.ReverseProxy{
+		Rewrite: func(pr *httputil.ProxyRequest) {
+			target := cfg.KubeRBACProxyInfo.UpstreamURL
+			pr.SetURL(target)
+			pr.Out.Host = target.Host
+			copyHeaderIfSet(pr.In, pr.Out, "X-Forwarded-For")
```
ibihim, line 220: Wouldn't it make sense to add the IP address of the client, if not already set?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#syntax

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
```diff
@@ -212,7 +212,15 @@ func Run(cfg *server.KubeRBACProxyConfig) error {
 		return fmt.Errorf("failed to setup an authorizer: %v", err)
 	}
 
-	proxy := httputil.NewSingleHostReverseProxy(cfg.KubeRBACProxyInfo.UpstreamURL)
+	proxy := &httputil.ReverseProxy{
+		Rewrite: func(pr *httputil.ProxyRequest) {
+			target := cfg.KubeRBACProxyInfo.UpstreamURL
+			pr.SetURL(target)
+			pr.Out.Host = target.Host
```
ibihim, line 219: Why do you set it, after rewrite removed it? To keep the same behavior as with the Director?

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
stlaz: `pr.SetXForwarded()` does that but only if `X-Forwarded-For` is already set

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
stlaz: Yes, I meant to keep the behavior the same.

ibihim: Yes, I assumed so, but I am wondering if this is some best practice like rewrite vs director :sweat_smile:

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
ibihim: oh, yes. I see. But why do we need to copy the `X-Forwarded-For` header? Does it get removed otherwise?

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
ibihim: Ok, I don't know why they are unsetting it, but it works with an empty request.Host, because of this: https://github.com/golang/go/blob/master/src/net/http/request.go#L605-L611. It will be automatically set, when the request is being written.

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
stlaz: See https://cs.opensource.google/go/go/+/refs/tags/go1.23.3:src/net/http/httputil/reverseproxy.go;l=417-424
and
https://cs.opensource.google/go/go/+/refs/tags/go1.23.3:src/net/http/httputil/reverseproxy.go;l=78-95

stlaz: Added unit tests.

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
stlaz: I followed https://pkg.go.dev/net/http/httputil#NewSingleHostReverseProxy

PR #325, opened 2024-11-25
e2e: don't pass client via globvar

stlaz:
Use the kube-native way to construct a client by using the KUBECONFIG env var. That was done in the tests anyway (see the change to the Makefile).

The current system promotes using globvars which is a malpractice that should be avoided. It's only confusing new contributors. Let's remove it and just use the KUBECONFIG variable as in the rest of the Kubernetes ecosystem.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,7 +97,7 @@ test-unit:
 	go test -v -race -count=1 $(PKGS)
 
 test-e2e:
-	go test -timeout 55m -v ./test/e2e/ $(TEST_RUN_ARGS) --kubeconfig=$(KUBECONFIG)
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/oklog/run v1.1.0
 	github.com/spf13/cobra v1.8.1
 	github.com/spf13/pflag v1.0.5
+	github.com/stretchr/testify v1.9.0
```
ibihim, line 11: Oh dear God, please no no no no no.

stlaz: You should explain.

Issue #318, opened 2024-11-12
(6) Wrapper to highlight tokens being audience agnostic link.


ibihim:
## What

We should use a OIDC authenticator to highlight, that it is audience agnostic. There is one upstream.

[link](https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-8796db11a6b137095413f7a7109dd01393e63d425454fc04647b98d2b8c0115cR183)

stlaz:
The current Kubernetes OIDC authenticator that's used in this case seems to be handling audiences and that's likely what we want in these cases - to enforce specific token audience for the KRP upstream. Isn't that the case?

PR #329, opened 2024-11-28
explicitly require OIDC audience for the OIDC authenticator

stlaz:
Fixes: https://github.com/brancz/kube-rbac-proxy/issues/318

### DIFF
--- a/cmd/kube-rbac-proxy/app/options/oidcoptions.go
+++ b/cmd/kube-rbac-proxy/app/options/oidcoptions.go
@@ -17,6 +17,8 @@ limitations under the License.
 package options
 
 import (
+	"fmt"
+
 	"github.com/brancz/kube-rbac-proxy/pkg/authn"
 	"github.com/brancz/kube-rbac-proxy/pkg/server"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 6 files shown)


stlaz:
@enj I don't think we want the OIDC authenticator to be audience agnostic in this case, instead we should require specific audience.
I would expect audience agnosticism to be required only in the case of the delegated authenticator.

enj:
> @enj I don't think we want the OIDC authenticator to be audience agnostic in this case, instead we should require specific audience. I would expect audience agnosticism to be required only in the case of the delegated authenticator.

@stlaz https://github.com/brancz/kube-rbac-proxy/pull/343#issuecomment-2549688353

cmd/kube-rbac-proxy/app/options/oidcoptions.go
```diff
@@ -41,6 +43,14 @@ func (o *OIDCOptions) AddFlags(flagset *pflag.FlagSet) {
 
 func (o *OIDCOptions) Validate() []error {
 	var errs []error
+	if len(o.IssuerURL) == 0 {
+		return errs
+	}
+
+	if len(o.RequiredAudience) == 0 {
+		errs = append(errs, fmt.Errorf("oidc-required-audience must be set when `oidc-issuer` is configured"))
+	}
```
enj, line 52: This should be included in #343

Issue #314, opened 2024-11-12
`transport.ForceAttemptHTTP2 = false` only in cert-cases?


ibihim:
## Question

Is it intended that `transport.ForceAttemptHTTP2` is set to `false` only if `if len(upstreamCAPath) > 0` or `if len(upstreamClientCertPath) > 0` are true?

[link](https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-d84eac55294d8f540a660bfcb4043f7c4e1f76c8bccbb91c129e8e4992dbafe7R95

PR #338, opened 2024-12-02
allow HTTPv1 for upstream connections unconditionally

stlaz:
We don't know the upstream HTTP version, there's likely no reason to force HTTP2 ever.

Fixes https://github.com/brancz/kube-rbac-proxy/issues/314


ibihim:
Makes sense. HTTP/2 tests didn't fail too.

Issue #333, opened 2024-11-29
(40) Not used? link


ibihim:

ShazaAldawamneh:
I can take this issue . @ibihim

PR #339, opened 2024-12-02
[WIP]: remove unused link, groups from userConfig in static.go

ShazaAldawamneh:
this ticket to resolve issue: [(40) Not used? link #333
](https://github.com/brancz/kube-rbac-proxy/issues/333 )

### DIFF
--- a/pkg/authorization/static/static.go
+++ b/pkg/authorization/static/static.go
@@ -38,8 +38,7 @@ type StaticAuthorizationConfig struct {
 }
 
 type UserConfig struct {
-	Name   string   `json:"name,omitempty"`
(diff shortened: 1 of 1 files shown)


stlaz:
What's the "unused link"?
What if, instead of a single user, I would like to allow a group of users to access the proxied service?

ShazaAldawamneh:
It doesn’t seem like the code uses “groups” in the way you’re suggesting, but I’m not sure if it could be utilized that way, as you mentioned. How can we determine if it’s necessary to support groups, or if the current implementation is sufficient? @stlaz

stlaz:
Please consider supporting groups instead of just users.

Issue #330, opened 2024-11-29
(33) Several header entries for groups, such that group separator can be used in group name link.


ibihim:

kramaranya:
I would like to take this issue @ibihim

PR #341, opened 2024-12-03
Put groups into separate header entries

kramaranya:
Add groups as separate header entries instead of concatenating them with a separator and update the corresponding unit test

Fixes: #330

### DIFF
--- a/pkg/authn/identityheaders/identityheaders.go
+++ b/pkg/authn/identityheaders/identityheaders.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	"k8s.io/apiserver/pkg/endpoints/request"
+	"k8s.io/klog/v2"
 )
 
 // AuthnHeaderConfig contains authentication header settings which enable more information about the user identity to be sent to the upstream
@@ -52,7 +53,15 @@ func WithAuthHeaders(handler http.Handler, cfg *AuthnHeaderConfig) http.Handler 
 			// Seemingly well-known headers to tell the upstream about user's identity
 			// so that the upstream can achieve the original goal of delegating RBAC authn/authz to kube-rbac-proxy
 			req.Header.Set(cfg.UserFieldName, u.GetName())
-			req.Header.Set(cfg.GroupsFieldName, strings.Join(u.GetGroups(), cfg.GroupSeparator))
+
+			if cfg.GroupSeparator == "" {
+				for _, group := range u.GetGroups() {
+					req.Header.Add(cfg.GroupsFieldName, group)
+				}
+			} else {
+				filteredGroups := filterGroups(u.GetGroups(), cfg.GroupSeparator)
+				req.Header.Set(cfg.GroupsFieldName, strings.Join(filteredGroups, cfg.GroupSeparator))
+			}
 		}
 
 		handler.ServeHTTP(w, req)
@@ -61,4 +70,16 @@ func WithAuthHeaders(handler http.Handler, cfg *AuthnHeaderConfig) http.Handler 
 
 func HasIdentityHeadersEnabled(cfg *AuthnHeaderConfig) bool {
 	return len(cfg.GroupsFieldName) > 0 || len(cfg.UserFieldName) > 0
+}
+
+func filterGroups(groups []string, separator string) []string {
+	var validGroups []string
+	for _, group := range groups {
+		if strings.Contains(group, separator) {
+			klog.Infof("Dropping group %q because it contains the group separator %q", group, separator)
+			continue
+		}
+		validGroups = append(validGroups, group)
+	}
+	return validGroups
 }
--- a/pkg/authn/identityheaders/identityheaders_test.go
+++ b/pkg/authn/identityheaders/identityheaders_test.go
@@ -21,7 +21,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/http/httputil"
-	"strings"
+	"reflect"
 	"testing"
 
 	"k8s.io/apiserver/pkg/authentication/authenticator"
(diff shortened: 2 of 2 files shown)


pkg/authn/identityheaders/identityheaders.go
```diff
@@ -52,7 +50,11 @@ func WithAuthHeaders(handler http.Handler, cfg *AuthnHeaderConfig) http.Handler
 			// Seemingly well-known headers to tell the upstream about user's identity
 			// so that the upstream can achieve the original goal of delegating RBAC authn/authz to kube-rbac-proxy
 			req.Header.Set(cfg.UserFieldName, u.GetName())
-			req.Header.Set(cfg.GroupsFieldName, strings.Join(u.GetGroups(), cfg.GroupSeparator))
```
stlaz, line 55: Does this drop the `GroupSeparator`?

While I am personally in favour of using the HTTP-standard header separation, I believe the original review comment requested a bit of a different outcome: https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-1ca0b12c9f8ec6aec4fea4c5542fe1bb7156876c19d791d1fcf7cf66e62905aeR57-R61. This is to support services that expect different groups formatting.

I suggest:
1. by default, do the right thing and behave like a good HTTP client (=> what you have here)
2. if a group separator is set, do what the code does currently.

On top of the above, we also need to:
- figure out what to do with groups that contain the standard separator ("," according to https://www.rfc-editor.org/rfc/rfc9110.html#name-header-fields) in the standard case - what does the Golang code do in that case?
- drop the groups containing the group separator as the original comment suggests in the `groupSeparator != ""` case

kramaranya: Thank you for making that clear! I've updated the code to address it and updated the unit tests to cover these scenarios

pkg/authn/identityheaders/identityheaders.go
```diff
@@ -17,10 +17,8 @@ limitations under the License.
 package identityheaders
 
 import (
-	"net/http"
-	"strings"
-
 	"k8s.io/apiserver/pkg/endpoints/request"
+	"net/http"
```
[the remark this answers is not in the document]
stlaz: imports ordering used in this project:
```
stdlib

3rd party

kube

local
```

pkg/authn/identityheaders/identityheaders.go
```diff
@@ -52,7 +53,31 @@ func WithAuthHeaders(handler http.Handler, cfg *AuthnHeaderConfig) http.Handler
 			// Seemingly well-known headers to tell the upstream about user's identity
 			// so that the upstream can achieve the original goal of delegating RBAC authn/authz to kube-rbac-proxy
 			req.Header.Set(cfg.UserFieldName, u.GetName())
-			req.Header.Set(cfg.GroupsFieldName, strings.Join(u.GetGroups(), cfg.GroupSeparator))
+
+			if cfg.GroupSeparator == "" {
+				// Add each group as a separate header entry
+				for _, group := range u.GetGroups() {
+					if strings.Contains(group, ",") {
```
stlaz, line 60: How does Golang handle a comma in headers?

pkg/authn/identityheaders/identityheaders.go
```diff
@@ -52,7 +53,31 @@ func WithAuthHeaders(handler http.Handler, cfg *AuthnHeaderConfig) http.Handler
 			// Seemingly well-known headers to tell the upstream about user's identity
 			// so that the upstream can achieve the original goal of delegating RBAC authn/authz to kube-rbac-proxy
 			req.Header.Set(cfg.UserFieldName, u.GetName())
-			req.Header.Set(cfg.GroupsFieldName, strings.Join(u.GetGroups(), cfg.GroupSeparator))
+
+			if cfg.GroupSeparator == "" {
+				// Add each group as a separate header entry
+				for _, group := range u.GetGroups() {
```
stlaz, line 59: turn the groups dropping logic into a function and reuse it in both branches

pkg/authn/identityheaders/identityheaders.go
stlaz: https://goplay.tools/snippet/n-xUX0e0w5K

I think we might be ok leaving even groups with `,` in

kramaranya: Thanks a lot for digging into this!

From what I’ve found, go doesnt treat commas inside a single header as a special separator -- it just stores and sends the value as is and combines multiple values into a single header line if needed. Also there is no documentation stating that golang removes or parses commas in any special way, at least I couldn't find it.. 

So since your code shows that the header round-trips correctly -- I agree that we don’t need to drop groups that contain commas. It should be safe to leave them in.

pkg/authn/identityheaders/identityheaders.go
kramaranya: updated, thank you :)

pkg/authn/identityheaders/identityheaders.go
```diff
@@ -52,7 +53,16 @@ func WithAuthHeaders(handler http.Handler, cfg *AuthnHeaderConfig) http.Handler
...
+
+			filteredGroups := filterGroups(u.GetGroups(), cfg.GroupSeparator)
+
+			if cfg.GroupSeparator == "" {
+				for _, group := range filteredGroups {
+					req.Header.Add(cfg.GroupsFieldName, group)
+				}
+			} else {
+				req.Header.Set(cfg.GroupsFieldName, strings.Join(filteredGroups, cfg.GroupSeparator))
+			}
```
stlaz, line 65: nit: since we now know we don't have to filter groups when a separator is not set, I would just:
```golang
if cfg.GroupSeparator == "" {
	for _, group := range u.GetGroups() {
			req.Header.Add(cfg.GroupsFieldName, group)
	}
} else {
	filteredGroups := filterGroups(u.GetGroups(), cfg.GroupSeparator)
	req.Header.Set(cfg.GroupsFieldName, strings.Join(filteredGroups, cfg.GroupSeparator))
}
```

and remove the `if separator == ""` branch in `filterGroups()`

stlaz:
The changes look good to me. I'm going to ask @ibihim to back me with a second pair of eyes :slightly_smiling_face:

ibihim (approved):
/lgtm

PR #343, opened 2024-12-11
cmd/krp/app,pkg/authn: make OIDC token aud agnostic

ibihim:
## What

Make OIDC authenticator audience agnostic.

## Why

Mo said so. And I assume the point is that the aud in an ID token will be the client ID and the overwrites the classical meaning of an "audience".

### DIFF
--- a/cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
+++ b/cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
@@ -189,7 +189,11 @@ func Run(cfg *server.KubeRBACProxyConfig) error {
 	var authenticator authenticator.Request
 	// If OIDC configuration provided, use oidc authenticator
 	if cfg.KubeRBACProxyInfo.OIDC.IssuerURL != "" {
-		oidcAuthenticator, err := authn.NewOIDCAuthenticator(ctx, cfg.KubeRBACProxyInfo.OIDC)
+		oidcAuthenticator, err := authn.NewOIDCAuthenticator(
+			ctx,
+			cfg.KubeRBACProxyInfo.OIDC,
+			cfg.DelegatingAuthentication.APIAudiences,
+		)
 		if err != nil {
 			return fmt.Errorf("failed to instantiate OIDC authenticator: %w", err)
 		}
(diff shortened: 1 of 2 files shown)


stlaz:
I don't see a reason why we would want to allow OIDC authenticator responses to _not_ contain audiences.

ibihim:
I know what you mean. I am not sure as well, but this is how I understood it.

enj:
> I don't see a reason why we would want to allow OIDC authenticator responses to _not_ contain audiences.

That is not what "audience agnostic" means.  Audience agnostic means that the authenticator never returns an `authenticator.Response` with the `Audiences` field set.  This is always true for the [OIDC authenticator](https://github.com/kubernetes/kubernetes/blob/107be8feccfce7259b402ed75415309268744a24/staging/src/k8s.io/apiserver/plugin/pkg/authenticator/token/oidc/oidc.go#L804), meaning it is always audience agnostic.  See https://github.com/kubernetes/kubernetes/pull/87612 for the history around this.  By using `WrapAudienceAgnosticToken`, all we are doing is artificially setting the `authenticator.Response.Audiences` to match the audiences configured for the proxy, which is strictly more correct for the proxy because it only has one concept of an audience, whereas the Kube API server has both "API audiences" and "JWT audiences."  FWIW, I don't think any consumer of the `authenticator.Response` in this codebase cares either way.

enj:
Hm, [this part of WrapAudienceAgnosticToken](https://github.com/kubernetes/kubernetes/blob/e305c3398896b04c0d3b58d64531f8a87f685f68/staging/src/k8s.io/apiserver/pkg/authentication/authenticator/audagnostic.go#L26-L31) would make all of this a no-op unless something is also calling `WithAudiences`.

ibihim:
@enj, yes this is a programmatic representation of my confusion. Thank you for the links for understanding! :smile:

enj:
I think this all comes down to:

- How likely are we to accept tokens that somehow target the proxy with a different audience than the OIDC one?  This seems highly unlikely to me.
- Does any of the code that consumes the authentication result care about the audience, now or in the future?

My gut says that either way will be fine, so if you prefer to do #329 since it is simpler, I am fine with that.

PR #346, opened 2024-12-13
[wip] Consume rewrite parameters, except told otherwise

ibihim:
## What

If header or query params are consumed by kube-rbac-proxy, they will be cleaned up, except that the authz config states:

```
"insecurePassthrough": "true
```

## Why

If a proxy and upstream interpret parameter from the request, the probability is high that they do it differently, which invites CVEs.
In Prometheus use-cases, we want that behavior, such that if someone wants to access namespace default, a particular authorization config can created the expected SubjectAccessRequest to the apiserver.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,7 @@ authorization:
     subresource: <string>
     name: <string>
   rewrites:
+    insecurePassthrough: <bool>
     byQueryParameter:
       name: <string>
     byHttpHeader:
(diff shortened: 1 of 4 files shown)


stlaz:
Per our offline discussion, we should get rid of the query rewrites and stick with the header handling.

I'd suggest the switch for the handling logic to operate like a mode switch rather than the bool. We would have two modes:
1. TerminatingEvaluation
2. PassthroughEvaluation
or whatever names seem fitting, and document them properly, including the dangers of passthrough.

Issue #331, opened 2024-11-29
(25) Rename config file name to authz config filename as it is specific to authz link.


ibihim:
## What

Rename the Config to AuthzConfig.

## Why

The config file, doesn't configure KRP as a whole, but is specific to authorization rules.

## Reference

https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-5379bac64778825441ad4a223c319e73645905b9864419d08eb7b8d2db974cddR48-R50

ShazaAldawamneh:
I can take this issue . @ibihim

PR #349, opened 2024-12-13
Rename config file name to authz config filename as it is specific to authz

ShazaAldawamneh:
this ticket to resolve issue: [(25) Rename config file name to authz config filename as it is specific to authz](https://github.com/brancz/kube-rbac-proxy/issues/331)

### DIFF
--- a/cmd/kube-rbac-proxy/app/options/proxyoptions.go
+++ b/cmd/kube-rbac-proxy/app/options/proxyoptions.go
@@ -133,8 +133,8 @@ func (o *ProxyOptions) ApplyTo(krpInfo *server.KubeRBACProxyInfo, authInfo *serv
 		return fmt.Errorf("failed to setup transport for upstream: %w", err)
 	}
 
-	if configFileName := o.AuthzConfigFileName; len(configFileName) > 0 {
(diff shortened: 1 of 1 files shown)


cmd/kube-rbac-proxy/app/options/proxyoptions.go
```diff
@@ -133,7 +133,7 @@ func (o *ProxyOptions) ApplyTo(krpInfo *server.KubeRBACProxyInfo, authInfo *serv
 		return fmt.Errorf("failed to setup transport for upstream: %w", err)
 	}
 
-	if configFileName := o.ConfigFileName; len(configFileName) > 0 {
+	if configFileName := o.AuthzConfigFileName; len(configFileName) > 0 {
 		krpInfo.Authorization, err = parseAuthorizationConfigFile(configFileName)
```
ibihim, line 137: It is not necessarily part of the issue, but I don't think that this is necessary: `configFileName := o.AuthzConfigFileName`. Maybe we could drop `configFileName` and just use `o.AuthzConfigFileName` in `parseAuthorizationConfigFile`.

stlaz (approved):

Issue #334, opened 2024-11-29
(44) Use `sync.Pool` instead? link


ibihim:
## What

Consider using `sync.Pool` in `templateWithValue`.

## Why

Because this is being executed by go routines (requests) and sync.Pool might be a good solution for that.

## Reference

https://github.com/brancz/kube-rbac-proxy/pull/229/files#diff-c5321b4f3131d9d6f28032421cf6e9bb16827f1f887c808a380d6a0eb1bf5488R188

kramaranya:
I would like to take this issue @ibihim

PR #350, opened 2025-01-02
Reuse buffer with sync.Pool

kramaranya:
While investigating a task #334 to consider `sync.Pool` in our code, I discovered that currently, each invocation of `templateWithValue` creates a new `bytes.Buffer` and discards it. By switching to a `sync.Pool` to reuse and reset these buffers, we can cut down on allocations and reduce overhead.

Here are the benchmark results I gathered using [`benchstat`](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat) to compare `before_change.txt` (no `sync.Pool`) versus `after_change.txt` (using `sync.Pool`):

```
name                old time/op    new time/op    delta
TemplateWithValue-12    171.5ns ± 4%   145.4ns ± 0%  -15.19% (p=0.000 n=10)

name                old alloc/op   new alloc/op   delta
TemplateWithValue-12    192B ± 0%     80B ± 0%    -58.33% (p=0.000 n=10)

name                old allocs/op  new allocs/op  delta
TemplateWithValue-12      6.0 ± 0%     4.0 ± 0%   -33.33% (p=0.000 n=10)
```
Reusing the buffer with `sync.Pool` is better because it lowers overhead from creating new buffers on each call, reduces memory usage and helps the garbage collector by cutting down on short-lived objects:

- Time per operation improved by about 15%.
- Memory usage dropped by more than 50%.
- Number of allocations reduced from 6 to 4.

### DIFF
--- a/pkg/authorization/rewrite/attributes.go
+++ b/pkg/authorization/rewrite/attributes.go
@@ -18,6 +18,7 @@ package rewrite
 import (
 	"bytes"
 	"context"
+	"sync"
 	"text/template"
 
 	"k8s.io/apiserver/pkg/authorization/authorizer"
@@ -95,6 +96,8 @@ type TemplatedResourceAttributesGenerator struct {
 	resource    *template.Template
 	subresource *template.Template
 	name        *template.Template
+
+	bufPool *sync.Pool
 }
 
 var _ AttributesGenerator = &TemplatedResourceAttributesGenerator{}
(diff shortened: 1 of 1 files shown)


pkg/authorization/rewrite/attributes.go
```diff
@@ -147,11 +148,20 @@ func (r *TemplatedResourceAttributesGenerator) Generate(ctx context.Context, att
 	return attrs
 }
 
+var bufPool = sync.Pool{
```
ibihim, line 151: Maybe I am a burned child as I worked with JavaScript in the crazy days, but usually you avoid having a "global" variable.
As it is local to the package, it might be okaish, if we would rename it to something like "templateBufferPool".

A better solution would be to initiate this in the constructor and pass it down from the struct into the templateWithValue function. According to the docs, we must not copy it, so we would need to use a pointer.

kramaranya: Thanks, I've addressed your suggestion

stlaz (approved):

PR #351, opened 2025-01-07
CNTRLPLANE-63: Consider supporting groups instead of single users only and create their tests

ShazaAldawamneh:
this PR to cover issue: [333](https://github.com/brancz/kube-rbac-proxy/issues/333 
)

### DIFF
--- a/pkg/authorization/static/static.go
+++ b/pkg/authorization/static/static.go
@@ -60,17 +60,32 @@ func (saConfig StaticAuthorizationConfig) Matches(a authorizer.Attributes) bool 
 	isAllowed := func(staticConf string, requestVal string) bool {
 		if staticConf == "" {
 			return true
-		} else {
-			return staticConf == requestVal
 		}
+		return staticConf == requestVal
+	}
+	isGroupAllowed := func(configGroups, requestGroups []string) bool {
+		if len(configGroups) == 0 {
+			return true
+		}
+		for _, configGroup := range configGroups {
+			for _, requestGroup := range requestGroups {
+				if configGroup == requestGroup {
+					return true
+				}
+			}
+		}
+		return false
 	}
 
 	userName := ""
+	userGroups := []string{}
 	if a.GetUser() != nil {
 		userName = a.GetUser().GetName()
+		userGroups = a.GetUser().GetGroups()
 	}
 
 	if isAllowed(saConfig.User.Name, userName) &&
+		isGroupAllowed(saConfig.User.Groups, userGroups) &&
 		isAllowed(saConfig.Verb, a.GetVerb()) &&
 		isAllowed(saConfig.Namespace, a.GetNamespace()) &&
 		isAllowed(saConfig.APIGroup, a.GetAPIGroup()) &&
--- a/pkg/authorization/static/static_test.go
+++ b/pkg/authorization/static/static_test.go
@@ -134,6 +134,55 @@ func TestStaticAuthorizer(t *testing.T) {
 				authorizer.AttributesRecord{Verb: "get", Resource: "services", ResourceRequest: true},
 			},
 		},
+		{
+			name: "groupMatch",
+			config: []StaticAuthorizationConfig{
+				{User: UserConfig{Groups: []string{"admin", "editors"}}, Verb: "get", Resource: "namespaces", ResourceRequest: true},
+			},
+			shouldPass: []authorizer.Attributes{
+				authorizer.AttributesRecord{User: &user.DefaultInfo{Name: "user1", Groups: []string{"admin"}}, Verb: "get", Resource: "namespaces", ResourceRequest: true},
+				authorizer.AttributesRecord{User: &user.DefaultInfo{Name: "user2", Groups: []string{"editors"}}, Verb: "get", Resource: "namespaces", ResourceRequest: true},
+			},
+			shouldNoOpinion: []authorizer.Attributes{
+				// User with non-matching group
+				authorizer.AttributesRecord{User: &user.DefaultInfo{Name: "user3", Groups: []string{"viewers"}}, Verb: "get", Resource: "namespaces", ResourceRequest: true},
+				// User with no groups
+				authorizer.AttributesRecord{User: &user.DefaultInfo{Name: "user4"}, Verb: "get", Resource: "namespaces", ResourceRequest: true},
+			},
+		},
+		{
+			name: "groupAndUserMatch",
+			config: []StaticAuthorizationConfig{
+				{User: UserConfig{Name: "system:foo", Groups: []string{"admin"}}, Verb: "get", Resource: "pods", ResourceRequest: true},
+			},
+			shouldPass: []authorizer.Attributes{
+				authorizer.AttributesRecord{User: &user.DefaultInfo{Name: "system:foo", Groups: []string{"admin", "dev"}}, Verb: "get", Resource: "pods", ResourceRequest: true},
+			},
+			shouldNoOpinion: []authorizer.Attributes{
+				// User name matches, but group does not
+				authorizer.AttributesRecord{User: &user.DefaultInfo{Name: "system:foo", Groups: []string{"viewers"}}, Verb: "get", Resource: "pods", ResourceRequest: true},
+				// Group matches, but user name does not
+				authorizer.AttributesRecord{User: &user.DefaultInfo{Name: "system:bar", Groups: []string{"admin"}}, Verb: "get", Resource: "pods", ResourceRequest: true},
+				// Neither user name nor group matches
+				authorizer.AttributesRecord{User: &user.DefaultInfo{Name: "system:baz", Groups: []string{"viewers"}}, Verb: "get", Resource: "pods", ResourceRequest: true},
+			},
+		},
+		{
+			name: "groupWildcard",
+			config: []StaticAuthorizationConfig{
+				{User: UserConfig{Groups: []string{}}, Verb: "get", Resource: "services", ResourceRequest: true},
+			},
+			shouldPass: []authorizer.Attributes{
+				// Any group should be allowed as no groups are specified
+				authorizer.AttributesRecord{User: &user.DefaultInfo{Name: "user1", Groups: []string{"viewers"}}, Verb: "get", Resource: "services", ResourceRequest: true},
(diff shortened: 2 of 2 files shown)


stlaz:
This is an open source project, remove any references to trackers that are not in this repo. If they contain additional context, move it to the GitHub issue you are fixing.

pkg/authorization/static/static_test.go
```diff
@@ -134,6 +134,55 @@ func TestStaticAuthorizer(t *testing.T) {
 				authorizer.AttributesRecord{Verb: "get", Resource: "services", ResourceRequest: true},
 			},
 		},
+		{
+			name: "groupMatch",
+			config: []StaticAuthorizationConfig{
+				{User: UserConfig{Groups: []string{"admin", "editors"}}, Verb: "get", Resource: "namespaces", ResourceRequest: true},
```
ibihim, line 140: Nit: I know this is "idiomatic" to the previous test data, but I would prefer, having readable specs:

```suggestion
			config: []StaticAuthorizationConfig{{
			    User: UserConfig{
			        Groups: []string{"admin", "editors"}
			    },
			    Verb: "get",
			    Resource: "namespaces",
			    ResourceRequest: true,
			},
```

ibihim (commented):
good job

pkg/authorization/static/static.go
```diff
@@ -60,17 +60,32 @@ func (saConfig StaticAuthorizationConfig) Matches(a authorizer.Attributes) bool
...
-			return staticConf == requestVal
 		}
+		return staticConf == requestVal
+	}
+	isGroupAllowed := func(configGroups, requestGroups []string) bool {
+		if len(configGroups) == 0 {
+			return true
+		}
+		for _, configGroup := range configGroups {
+			for _, requestGroup := range requestGroups {
```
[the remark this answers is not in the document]
ibihim: Might be useful to add this as a comment, in case that someone tries to debug performance issues in the future :smile:

pkg/authorization/static/static.go
```diff
@@ -60,17 +60,33 @@ func (saConfig StaticAuthorizationConfig) Matches(a authorizer.Attributes) bool
...
-		} else {
-			return staticConf == requestVal
 		}
+		return staticConf == requestVal
+	}
+	isGroupAllowed := func(configGroups, requestGroups []string) bool {
+		if len(configGroups) == 0 {
+			return true
+		}
+		// O(n^2) is fine here as the groups are small. Optimize if n grows large.
```
stlaz, line 70: How do you plan to measure n growing large? I'd just use sets right away for the groups from config.

pkg/authorization/static/static.go
```diff
@@ -60,17 +60,33 @@ func (saConfig StaticAuthorizationConfig) Matches(a authorizer.Attributes) bool
...
 
 	userName := ""
+	userGroups := []string{}
 	if a.GetUser() != nil {
 		userName = a.GetUser().GetName()
+		userGroups = a.GetUser().GetGroups()
 	}
 
 	if isAllowed(saConfig.User.Name, userName) &&
+		isGroupAllowed(saConfig.User.Groups, userGroups) &&
```
stlaz, line 89: I am not sure I follow the use case. My expectation: "I want all users from group X to be able to proceed". However the implementation seems to require the user to be in certain groups beside it matching the username.

pkg/authorization/static/static.go
```diff
@@ -60,17 +60,34 @@ func (saConfig StaticAuthorizationConfig) Matches(a authorizer.Attributes) bool
...
 
 	userName := ""
+	userGroups := []string{}
 	if a.GetUser() != nil {
 		userName = a.GetUser().GetName()
+		userGroups = a.GetUser().GetGroups()
 	}
 
-	if isAllowed(saConfig.User.Name, userName) &&
+	if (saConfig.User.Name == "" || isAllowed(saConfig.User.Name, userName)) &&
```
ibihim, line 89: I think the previous solution was sufficient. If there is no config (empty string) for `user` or any other resourceAttribute, it defaults to true.

The unit tests show those permutations of user / no user, group / no group, allowed.

pkg/authorization/static/static.go
```diff
@@ -60,17 +60,34 @@ func (saConfig StaticAuthorizationConfig) Matches(a authorizer.Attributes) bool
...
-		} else {
-			return staticConf == requestVal
 		}
+		return staticConf == requestVal
+	}
+	isGroupAllowed := func(configGroups, requestGroups []string) bool {
+		if len(configGroups) == 0 {
+			return true
+		}
+		configGroupSet := make(map[string]struct{})
```
ibihim, line 70: A set is a great solution for this. You are always on the safe side that it scales up.

But the `configGroupSet` slice will never change, and we are initializing it on every request at least once.
We would need to move it out and initialize the `configGroupSet` in the constructor `NewStaticAuthorizer`.
The issue is, that the config given to the user to configure it is a slice of strings. This should stay as is, but for working with it we need to turn this into a Set of strings. So you would need to create as struct `StaticAuthorizationOptions` that contains a slice of strings and can be used by the user and you need to modify the existing `StaticAuthorizationConfig` to have a Set of strings.
If you create a Set and you don't use the k8s utility package and you do it by hand, it is also good to create `struct{}{}` as a global variable with a name like `noop` and reference it, that way you don't initialize it for every entry.

ShazaAldawamneh: > So you would need to create as struct StaticAuthorizationOptions that contains a slice of strings and can be used by the user

for this part, I already have a struct type called UserConfig is this not the same as the struct you're asking me to create ?

ibihim: The difference is, when is the Set being created: Once during initialization or on every request.

pkg/authorization/static/static.go
```diff
@@ -48,8 +49,15 @@ type staticAuthorizer struct {
...
-	for _, c := range config {
-		if c.ResourceRequest != (c.Path == "") {
+	for c := range config {
+		if config[c].User.Groups != nil {
+			groupSet := make(map[string]struct{})
+			for _, group := range config[c].User.Groups {
+				groupSet[group] = struct{}{}
+			}
+			config[c].User.GroupSet = groupSet
+		}
```
ibihim, line 59: Use [k8s.io/utils/set](https://github.com/kubernetes/utils/blob/24370beab75816ab1d2b33fba9e288ee667530dd/set/set.go#L28-L35):

```suggestion
		if config[i].User.Groups != nil {
			config[i].User.GroupSet = set.New(config[i].User.Groups...)
		}

```

pkg/authorization/static/static.go
```diff
@@ -38,8 +38,9 @@ type StaticAuthorizationConfig struct {
 }
 
 type UserConfig struct {
-	Name   string   `json:"name,omitempty"`
-	Groups []string `json:"groups,omitempty"`
+	Name     string   `json:"name,omitempty"`
+	Groups   []string `json:"groups,omitempty"`
+	GroupSet map[string]struct{}
```
ibihim, line 43: Use [k8s.io/utils/set](https://github.com/kubernetes/utils/blob/24370beab75816ab1d2b33fba9e288ee667530dd/set/set.go#L28-L35):

```suggestion
	GroupSet set.Set[string]
```

pkg/authorization/static/static.go
```diff
@@ -48,8 +50,12 @@ type staticAuthorizer struct {
 
 // NewStaticAuthorizer creates an authorizer for static SubjectAccessReviews
 func NewStaticAuthorizer(config []StaticAuthorizationConfig) (*staticAuthorizer, error) {
-	for _, c := range config {
-		if c.ResourceRequest != (c.Path == "") {
+	for c := range config {
```
stlaz, line 53: Since we need some internal representation of the config anyway, it might be easier to implement this part as a `unionauthorizer.New([]staticAuthorizer{...})` ([union authorizer constructor](https://github.com/kubernetes/kubernetes/blob/a0a226d158474a356816eee9adb786b7c24f9828/staging/src/k8s.io/apiserver/pkg/authorization/union/union.go#L40-L42)) where each staticAuthorizer in the above mentioned slice represents each element from `[]StaticAuthorizationConfig` here.

That way `func (saConfig StaticAuthorizationConfig) Matches(a authorizer.Attributes) bool`
changes into
`func (sa staticAuthorizer) Authorize(ctx context.Context, a authorizer.Attributes) (authorized authorizer.Decision, reason string, err error)`.
Each `return true` changes into `authorizer.DecisionAllow` and each `return false` into `authorizer.DecisionNoOpinion`.

 The constructor might then look something like this:
```go
func NewStaticAuthorizer(config []StaticAuthorizationConfig) (authorizer.Authorizer, error) {
   var authorizers []staticAuthorizer
   for _, c := range config {
       authz, err := newStaticAuthorizers(&c)
       // handle error
       authorizers = append(authorizers, authz)
   }
   return unionauthorizer.New(authorizers...)
}
```

WDYT?

cc @ibihim

pkg/authorization/static/static.go
```diff
@@ -60,17 +66,30 @@ func (saConfig StaticAuthorizationConfig) Matches(a authorizer.Attributes) bool
...
-			return staticConf == requestVal
 		}
+		return staticConf == requestVal
+	}
+	isGroupAllowed := func(requestGroups []string) bool {
+		if len(saConfig.User.GroupSet) == 0 {
+			return true
+		}
+		for _, group := range requestGroups {
+			if _, exists := saConfig.User.GroupSet[group]; exists {
```
stlaz, line 77: use the `set.Has()` method

pkg/authorization/static/static.go
ibihim: lol. Yes, I agree. This is what I tried to convey. We need an internal and an external representation. The internal one has a Set, the other has a slice of groups potentially.

WRT the I have no strong opinions. It could make sense to do that.

ibihim:
@stlaz, I tried to split options and config, but this looks pretty intense:

https://github.com/brancz/kube-rbac-proxy/commit/f5a229f56fef5a8a98d7eb4721ccaad3beef0309

It looks like we are adding a lot of complexity to keep the formalities correct. WDYT?

PR #357, opened 2025-02-04
WIP: Add manual opt-in for legacy service account tokens

### LINKED ISSUE #336: (28) Force a manual opt-in for legacy SA token? link

kramaranya:
Fixes #336


kramaranya:
@stlaz @ibihim could you please review this pr?

ibihim (commented):
The unit tests might not have been necessary, but fixing the e2e-tests would be crucial :) 

You also need to regenerate the README.md with the `make generate`.

ibihim:
@kramaranya, I think we have an audience token test, right? we might be able to drop it, if we do not have any tests without audience set, right? Maybe we could have a negative test case then, that the server fails if it isn't set?

stlaz (changes requested):
Will legacy SA tokens be denied when the new flag is unset but audiences are set?
Will they be accepted if both are set?

Please add e2e tests for all possible scenarios, I don't think I've found legacy SA tokens creation in the e2e tests.

cmd/kube-rbac-proxy/app/options/proxyoptions_test.go
```diff
@@ -125,3 +126,93 @@ func Test_parseAuthorizationConfigFile(t *testing.T) {
...
+		UpstreamClientCertFile          string
+		UpstreamClientKeyFile           string
+		UpstreamHeader                  *identityheaders.AuthnHeaderConfig
+		AuthzConfigFileName             string
+		AllowPaths                      []string
+		IgnorePaths                     []string
+		ProxyEndpointsPort              int
+		TokenAudiences                  []string
+		AllowLegacyServiceAccountTokens bool
+		DisableHTTP2Serving             bool
```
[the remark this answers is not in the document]
kramaranya: Sure, I've added additional test cases to cover those fields :)

test/e2e/allowpaths/deployment.yaml
```diff
@@ -22,6 +22,7 @@ spec:
             - "--proxy-endpoints-port=8643"
             - "--upstream=http://127.0.0.1:8081/"
             - "--allow-paths=/metrics,/api/v1/label/*"
+            - "--allow-legacy-serviceaccount-tokens=true"
```
stlaz, line 25: Please also add tests for scenarios that actually use the legacy token according to https://github.com/brancz/kube-rbac-proxy/pull/357#pullrequestreview-2623810842

kramaranya: oops completely forgot about this comment. will do, thanks for the reminder!

test/e2e/basics.go
```diff
@@ -519,3 +519,127 @@ func testIgnorePaths(client kubernetes.Interface) kubetest.TestSuite {
...
+				kubetest.ServiceIsReady(
+					client,
+					"kube-rbac-proxy",
+				),
+			),
+			Then: kubetest.Actions(
+				kubetest.ClientSucceeds(
+					client,
+					legacyCommand,
+					&kubetest.RunOptions{TokenAudience: "kube-rbac-proxy"},
```
stlaz, line 601: we don't need the options in this case, right?

test/e2e/basics.go
```diff
@@ -519,3 +519,127 @@ func testIgnorePaths(client kubernetes.Interface) kubetest.TestSuite {
...
+				kubetest.ServiceIsReady(
+					client,
+					"kube-rbac-proxy",
+				),
+			),
+			Then: kubetest.Actions(
+				kubetest.ClientSucceeds(
+					client,
+					command,
+					&kubetest.RunOptions{TokenAudience: ""},
```
stlaz, line 640: This is equal to it being nil, right?

test/e2e/basics.go
```diff
@@ -519,3 +519,127 @@ func testIgnorePaths(client kubernetes.Interface) kubetest.TestSuite {
 		}.Run(t)
 	}
 }
+
+func testLegacyTokenFlag(client kubernetes.Interface) kubetest.TestSuite {
+	return func(t *testing.T) {
+		legacyCommand := `curl --connect-timeout 5 -v -s -k --fail -H "Authorization: Bearer $(cat /var/run/secrets/tokens/requestedtoken)" https://kube-rbac-proxy.default.svc.cluster.local:8443/metrics`
```
stlaz, line 525: `/var/run/secrets/tokens/requestedtoken` does not actually house the legacy tokens. The below code appears to be creating such tokens:

https://github.com/brancz/kube-rbac-proxy/blob/670377fd78875cfcfc9a1c42e0b086da8bb81043/test/kubetest/kubernetes.go#L524-L544

We may actually want to rename the directory from `/var/run/secrets/tokens` to something else, like `/var/run/projected/tokens`, I personally find it a bit confusing.

You'll have to follow https://kubernetes.io/docs/concepts/configuration/secret/#serviceaccount-token-secrets to create a secret along with your other manifests. Use a different directory than for the short-lived SA tokens so that we can distinguish these two in tests.

test/e2e/basics.go
```diff
@@ -519,3 +519,127 @@ func testIgnorePaths(client kubernetes.Interface) kubetest.TestSuite {
...
+		command := `curl --connect-timeout 5 -v -s -k --fail -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" https://kube-rbac-proxy.default.svc.cluster.local:8443/metrics`
```
stlaz, line 526: I think this command should actually have `/var/run/secrets/tokens/requestedtoken` as an input for the token

stlaz (changes requested):
We'll have to create the legacy SA token separately

test/e2e/basics.go
kramaranya: The [LegacyAllowedNoAudience](https://github.com/kramaranya/kube-rbac-proxy/blob/legacy-sa-opt-in/test/e2e/basics.go#L569-L572) test checks a scenario with no audience set. But `/var/run/secrets/tokens/requestedtoken` sets an audience, so it wouldn't match our no-audience case. 
https://github.com/brancz/kube-rbac-proxy/blob/670377fd78875cfcfc9a1c42e0b086da8bb81043/test/kubetest/kubernetes.go#L524-L532
That's why I use the default token `/var/run/secrets/kubernetes.io/serviceaccount/token` instead.

stlaz: The `/var/run/secrets/kubernetes.io/serviceaccount/token` will also contain an audience.

Issue #359, opened 2025-02-12
Looks like CVE-2024-45337 reported on 0.18.2


Alinkerror:
The security scans reported  CVE-2024-45337 on latest branch for golang.org/x/crypto. The suggested upgrade is to move to version > 0.31.0. Any timeline when it would be fixed?

ibihim:
Yes, this CVE looks pretty bad, in general.
But we don't use Public Key cryptography, except in our test, to be specific.

Nevertheless, we didn't do a release for some time, we could start the process of releasing a new one.

Thanks for highlighting this issue.

stlaz:
Note that KRP is not affected by the CVE

ibihim:
I assumed so, but I was too lazy to verify that we are not using `ServerConfig.PublicKeyCallback` indirectly and as a release was due anyway... 🤷

Alinkerror:
There is CVE-2024-45338 raised for golang.org/x/net too which is being consume in KRP. An update to the latest crypto and x/net would be good :)

Alinkerror:
@ibihim - Any idea when the new release with the bump up dedenencies would be available?

PR #360, opened 2025-02-14
Candidate v0.19.0

ibihim:
## What

Create release v0.19.0

## Why

- CVE in deps
- Didn't release in a while
- And we are having a new feature

## Notes

Ref: https://github.com/brancz/kube-rbac-proxy/issues/359


stlaz:
lgtm, let's wait for the test report

go.mod
```diff
@@ -1,25 +1,28 @@
 module github.com/brancz/kube-rbac-proxy
 
-go 1.23
+go 1.23.0
+
+toolchain go1.23.6
```
stlaz, line 5: I think we may not want to set this

stlaz (changes requested):

go.mod
ibihim: Yes, this was automagically set. I think it is fine as it indicates the minimal viable Go version in the sense that it might use features from v1.23.0+

stlaz (approved):

PR #362, opened 2025-03-07
Update outdated GitHub Actions

lucacome:
Updates the Actions to the latest version available. This among other things removes the warnings in the runs.

The `setup-go` action now accepts `stable` as a version to always use the latest stable version.

Also updated `.golangci.yaml` to remove the warnings about deprecated configuration options.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,70 +2,107 @@ name: Go
 on: [push, pull_request]
 env:
   QUAY_PATH: quay.io/brancz/kube-rbac-proxy
-  go-version: '1.23'
   kind-version: 'v0.25.0'
+
+defaults:
+  run:
+    shell: bash
+
+concurrency:
+  group: ${{ github.ref_name }}-go
+  cancel-in-progress: true
+
 jobs:
   check-license:
     runs-on: ubuntu-latest
     name: Check license
     steps:
-    - uses: actions/checkout@v2
-    - run: make check-license
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Check license
+      run: make check-license
+
   generate:
     runs-on: ubuntu-latest
     name: Generate
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-go@v2
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Setup Golang Environment
+      uses: actions/setup-go@v5
       with:
-        go-version: ${{ env.go-version }}
-    - run: make generate && git diff --exit-code
+        go-version: stable
+
+    - name: Generate
+      run: make generate && git diff --exit-code
+
   lint:
     runs-on: ubuntu-latest
     name: Lint
     steps:
-    - uses: actions/checkout@v2
-    - name: golangci-lint
-      uses: golangci/golangci-lint-action@v2
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Lint Go
+      uses: golangci/golangci-lint-action@v6
       with:
         version: latest
         args: --timeout=5m
+
   build:
     runs-on: ubuntu-latest
     name: Build
     steps:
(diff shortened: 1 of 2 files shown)


ibihim:
Oh, that is a cool contribution. I will take a look, but first I will take a look at the bug introduced with the last version.

lucacome:
Is there anything I need to do to get this merged?

ibihim:
I don't think so, just patience 😅

I was waiting for @stlaz, but he said that it is fine to merge. But there is no hurry yet as there are a couple of other PRs to merge before the next release.

ibihim:
@lucacome , actually I just watched a best practice in a talk at KubeCon London:

Could you pin the full commit hash of the current checkout v4 ([11bd71901bbe5b1630ceea73d27597364c9af683](https://github.com/actions/checkout/commit/11bd71901bbe5b1630ceea73d27597364c9af683))? It is considered more secure, than following the tag.

lucacome:
@ibihim I usually have the SHAs for all the Actions in all my repos, but I also add something like dependabot or renovate to keep the dependencies up to date. 
I feel like pinning a dependency that will never get updated again might be worse than having a major tag. What do you think?

.github/workflows/build.yml
```diff
@@ -1,71 +1,108 @@
...
+      uses: actions/setup-go@0aaccfd150d50ccaeb58ebd88d36e91967a5f35b # v5.4.0
+      with:
+        go-version: stable
+
     - name: Start kind & create cluster
-      uses: engineerd/setup-kind@v0.5.0
+      uses: engineerd/setup-kind@71e45b960fc8dd50b4aeabf6eb6ef2ca0920b4c1 # v0.6.2
       with:
-        version: ${{ env.kind-version }}
+        version: "v0.27.0"
```
ibihim, line 99: It is better to have this at the top, so one doesn't need to scroll through the file to bump the version

ibihim:
Can you squash also the commits into one or two (e.g. one for the github action and one for the other stuff)? :smile:

.github/workflows/build.yml
lucacome: I reverted it. If we decide to use renovate we can keep that up to date automatically too.

lucacome:
> I agree, we will need to change that :) If you have any best practices to share, you are welcome, otherwise I would make it part of the release cycle, to not only bump go deps, but also GitHubActions.

I would add renovate. I opened a PR a while ago with dependabot (because you can just add the file to enable it) and it was rejected tho.. 😅

lucacome:
By the way, I was also planning to add the build step as a GitHub Action instead of the script after this one is merged.

.github/workflows/build.yml
```diff
@@ -1,71 +1,110 @@
...
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+    - name: Setup Golang Environment
+      uses: actions/setup-go@0aaccfd150d50ccaeb58ebd88d36e91967a5f35b # v5.4.0
+      with:
+        go-version: stable
+
     - name: Start kind & create cluster
-      uses: engineerd/setup-kind@v0.5.0
+      uses: helm/kind-action@a1b0e391336a6ee6713a0583f8c6240d70863de3 # v1.12.0
```
stlaz, line 98: I haven't followed the PR from the beginning, was this change explained somewhere?

.github/workflows/build.yml
```diff
@@ -1,71 +1,110 @@
 name: Go
+
 on: [push, pull_request]
+
 env:
   QUAY_PATH: quay.io/brancz/kube-rbac-proxy
-  go-version: '1.23'
```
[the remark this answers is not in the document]
ibihim: hm, ok. Deterministic is better than probabilistic.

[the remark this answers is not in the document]
stlaz: x.y like we had before should be fine

PR #364, opened 2025-03-13
[wip] issue 356 drop or not to drop attrs

ibihim:
## What

- Remove non resource attributes generator.

## Why

- We might not need it. Normal requests already have the given form of only "path", "verb" and "user".
- It might be more secure, as we don't mask other attributes.

## Notes

- It is theoretically a regression.


ibihim:
@ibihim add log output before change, to show an exemplary resource attribute for that case.

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
```diff
@@ -272,38 +271,20 @@ func Run(cfg *server.KubeRBACProxyConfig) error {
...
+
 	// Static authorization authorizes against a static file is ran before the SubjectAccessReview.
 	if krbInfo.Authorization.Static != nil {
 		staticAuthorizer, err := static.NewStaticAuthorizer(krbInfo.Authorization.Static)
 		if err != nil {
 			return nil, fmt.Errorf("failed to create static authorizer: %w", err)
 		}
 
-		delegated = append(delegated, staticAuthorizer)
+		authz = union.New(staticAuthorizer, authz)
```
stlaz, line 286: This is an odd change. Can you please write full unit test coverage of the original code (so a unit tests commit comes first) and then start moving authorizers around?

stlaz (changes requested):

PR #365, opened 2025-03-14
authn: test against (non) nil CA

ibihim:
## What

Add unit tests to test the OIDC logic.

## Why

There was an issue before and a comment isn't good enough to protect someone in the future from breaking it :smile:

### DIFF
--- a/pkg/authn/oidc_test.go
+++ b/pkg/authn/oidc_test.go
@@ -1,0 +1,127 @@
+/*
+Copyright 2025 the kube-rbac-proxy maintainers All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


pkg/authn/oidc_test.go
```diff
@@ -0,0 +1,127 @@
...
+			w.Header().Set("Content-Type", "application/json")
+			fmt.Fprint(w, `{"keys": []}`)
+		default:
+			http.NotFound(w, r)
+		}
+	}))
+	return ts
+}
+
+func TestNewOIDCAuthenticator(t *testing.T) {
```
stlaz, line 48: This only tests the CAs used in the authenticator.

Either rename the test or dedicate some time to turn this into table tests for the whole authenticator config (the latter is preferred).

pkg/authn/oidc_test.go
```diff
@@ -0,0 +1,127 @@
...
+	// Create a dummy OIDC issuer (HTTPS).
+	ts := newTestOIDCIssuer()
+	defer ts.Close()
+
+	// Override the default transport to skip TLS verification.
+	// This is needed because ts uses a self-signed certificate.
+	// This should be legit for testing purpose, even with parallel test cases run at the same time.
```
stlaz, line 55: I don't think this is legit.

stlaz (changes requested):

ibihim:
not worth the effort

PR #366, opened 2025-03-20
make update go deps

ibihim:
## What

- Update go deps
- Remove deprecated flag

## Why

- GitHub bot created a PR to fix some CVE that doesn't apply to us, but it makes uses less worried.
- Flag is no-op and will be removed someday in the future.


Makefile
```diff
@@ -52,7 +52,7 @@ build: clean $(OUT_DIR)/$(PROGRAM_NAME)
 
 update-go-deps:
 	@for m in $$(go list -mod=readonly -m -f '{{ if and (not .Indirect) (not .Main)}}{{.Path}}{{end}}' all); do \
-		go get -d $$m; \
+		go get $$m; \
```
stlaz, line 55: given the name of the target, it seems like it was meant to be
```suggestion
		go get -u $$m; \
```
instead

cdash415:
@ibihim Any ETA on the next release with this change? this will probably the fix for [CVE-2025-22870]

ibihim:
@cdash415, I am aware of it now. I will work on a release.

cdash415:
Thanks for acknowledging and updating it. Hopefully will get the fix after merge

PR #377, opened 2025-05-21
test/e2e: Authz tests in combination

ibihim:
## What

Add more authorization tests to test the authorization chain, if different authorization methods are set.

## Why

Those tests are missing and make it hard to guarantee the same behavior in the sig-auth-acceptance branch.


test/e2e/authz-templated-query-rewrite-static/clusterRole.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
stlaz, line 1: revert all of the file moves, like [test/e2e/static-auth/serviceAccount.yaml → test/e2e/authz-static/serviceAccount.yaml](https://github.com/brancz/kube-rbac-proxy/pull/377/commits/b8d2d7efc658877f0f1178a062a96f21e0c57fc9#diff-421e293188e06df7e377bb2fb4ce47028e002e841aa725b4b3b89c3c3d224dcd)
This is really hard to review otherwise and the benefit of the move is minimal

test/e2e/main_test.go
```diff
@@ -38,18 +38,18 @@ func Test(t *testing.T) {
...
-		"AllowPath":          testAllowPathsRegexp(client),
-		"IgnorePath":         testIgnorePaths(client),
-		"TLS":                testTLS(client),
-		"StaticAuthorizer":   testStaticAuthorizer(client),
-		"HTTP2":              testHTTP2(client),
-		"Flags":              testFlags(client),
-		"TokenMasking":       testTokenMasking(client),
-		"CombinedAuthorizer": testCombinedAuthorizer(client),
+		// "Basics":             testBasics(client),
+		// "H2CUpstream":        testH2CUpstream(client),
```
stlaz: wat

test/e2e/static_authorizer.go
```diff
@@ -39,19 +39,48 @@ func testStaticAuthorizer(client kubernetes.Interface) kubetest.TestSuite {
...
 						fmt.Sprintf(command, "/metrics"),
-						nil,
+						&kubetest.RunOptions{
+							ServiceAccount: "client-with-static-match",
+						},
+					),
+				),
+			},
+			{
+				name: "static/forbidden -> rbac/granted",
```
stlaz, line 62: We need good test names given the test framework makes the tests super cryptic on their own. I am having a hard time understanding what this might mean.

PR #441, opened 2026-07-03
server: demote TCP-reset TLS handshake errors to verbose logging

andreacv98:
## What

Introduce `tlsHandshakeErrorFilter` as `http.Server.ErrorLog` for both TLS listeners. Messages matching both `"http: TLS handshake error"` and `"connection reset by peer"` are demoted to `V(4)`, remaining accessible under `-v=4`. All other TLS handshake errors (expired certificates, unknown authorities, unsupported protocol versions) are unaffected.

## Why

Load balancers such as HAProxy perform periodic TCP-level health checks by opening a connection and immediately closing it with a TCP RST before the TLS handshake completes. This generates a continuous stream of `http: TLS handshake error ... connection reset by peer` log messages at the default level with no functional impact.

Verified on an OpenShift cluster with HAProxy as ingress controller:  openshift/kube-rbac-proxy#145

### DIFF
--- a/cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
+++ b/cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
@@ -22,6 +22,7 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"log"
 	"net"
 	"net/http"
 	"net/http/httputil"
@@ -322,6 +323,7 @@ func Run(cfg *completedProxyRunOptions) error {
 			srv := &http.Server{
 				Handler:   mux,
 				TLSConfig: &tls.Config{},
+				ErrorLog:  log.New(&tlsHandshakeErrorFilter{}, "", 0),
 			}
 
 			if cfg.tls.CertFile == "" && cfg.tls.KeyFile == "" {
@@ -412,6 +414,7 @@ func Run(cfg *completedProxyRunOptions) error {
 				proxyEndpointsSrv := &http.Server{
 					Handler:   proxyEndpointsMux,
 					TLSConfig: srv.TLSConfig.Clone(),
+					ErrorLog:  log.New(&tlsHandshakeErrorFilter{}, "", 0),
 				}
 
 				if cfg.http2Disable {
--- a/cmd/kube-rbac-proxy/app/tls_error_filter.go
+++ b/cmd/kube-rbac-proxy/app/tls_error_filter.go
@@ -1,0 +1,52 @@
+/*
+Copyright 2026 the kube-rbac-proxy maintainers. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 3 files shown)


stlaz:
These are important signals, we'll stick with the default logging levels used by everyone else.

It's a bit odd that KRP would be the only project in OpenShift that behaves like this behind HAProxy.

vwalek:
Hello @stlaz

I would like to reopen the discussion regarding moving TLS handshake error logging to a higher verbosity level.
An argument for this request is that kube-apiserver already offloads handshake errors to log level 5 in its http.Server with ErrorLog implementation (see references below).

https://github.com/kubernetes/apiserver/commit/7f8013437d6b9c7e06f1fad091d5623b42411392 
https://github.com/kubernetes/apiserver/blob/master/pkg/server/secure_serving.go#L209-L212 
https://github.com/kubernetes/apiserver/blob/master/pkg/server/secure_serving.go#L299 

The kube apiserver will handle all the handshake errors and will log them under loglevel=5.
```
// tlsHandshakeErrorWriter writes TLS handshake errors to klog with
// trace level - V(5), to avoid flooding of tls handshake errors.
type tlsHandshakeErrorWriter struct {
	out io.Writer
}

const tlsHandshakeErrorPrefix = "http: TLS handshake error"

func (w *tlsHandshakeErrorWriter) Write(p []byte) (int, error) {
	if strings.Contains(string(p), tlsHandshakeErrorPrefix) {
		klog.V(5).Info(string(p))
		metrics.TLSHandshakeErrors.Inc()
		return len(p), nil
	}

	// for non tls handshake error, log it as usual
	return w.out.Write(p)
}
```


Based on my experience, working alongside system admins and customers, these error logs are rarely helpful and often do not provide much value during the troubleshooting.

Historic logs usually don't explain why the issue occurred, so in many cases the admin has to reproduce the issue and wait for it to happen again. Considering that a TCP health check generates 12 messages per minute (a 5s interval), roughly ~700 per hour: if a client fails once per hour, pinpointing that single occurrence is a needle in a haystack. Raising the log level and then reproducing the issue,is the preferable approach. I did check the error messages in my lab cluster, and the error message was found in 8 different kube-rbac-proxy containers, with a total of ~3,500 occurrences just within ~2 hours of the cluster creation.

At the standard log level, "connection reset by peer" conveys almost nothing beyond the fact that the connection failed at the TCP layer of the networking stack. When a client connects to the API endpoint and the connection fails at the TCP level, the client sees a generic "Something went wrong" or "Connection failed" error. 
Setting the TLS handshake error to log level 5 provides a feasible resolution to the issue, since verbosity can be increased on demand whenever analysis is needed.

The code change in this PR does define that only connections reset by peer should be logged under level 5, but I would recommend to follow the same implementation as in kube apiserver.
```
func isTCPResetHandshakeError(msg string) bool {
	return strings.Contains(msg, "http: TLS handshake error") &&
		strings.Contains(msg, "connection reset by peer")
}
```

Please let me know what you think.

Additionally, the request to log the handshake errors to higher verbosity directly in the http package was discussed in the issue [GH#26918](https://github.com/golang/go/issues/26918). It was closed with a [proposal](https://github.com/golang/go/issues/26918#issuecomment-412171562) to set an ErrorLog and filter the errors.

PR #443, opened 2026-07-05
proxy: sanitize headers and normalize paths before upstream

yahlifried:
Strip Authorization and identity headers (X-Remote-User, X-Remote-Groups, X-Remote-Extra-*) from requests before forwarding to the upstream on the ignore-paths branch. The upstream process has no legitimate need for caller credentials on unauthenticated paths such as health checks.

Ensure WithAuthHeaders always strips inbound identity headers regardless of whether --auth-header-fields-enabled is set. When enabled, the proxy sets these from the authenticated user context; when disabled, they should not pass through from the caller. Identity headers reaching the upstream must be proxy-authoritative.

Normalize URL paths via path.Clean before matching against --ignore-paths and --allow-paths patterns. Reject non-canonical paths (containing .., //, trailing slashes) with 400 Bad Request to ensure consistent matching behavior.

### DIFF
--- a/cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
+++ b/cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
@@ -283,9 +283,16 @@ func Run(cfg *completedProxyRunOptions) error {
 	}
 
 	handler := http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		cleaned := path.Clean(req.URL.Path)
+		if cleaned != req.URL.Path {
+			http.Error(w, "Bad Request", http.StatusBadRequest)
+			return
+		}
+
 		ignorePathFound := false
 		for _, pathIgnored := range cfg.ignorePaths {
-			ignorePathFound, err = path.Match(pathIgnored, req.URL.Path)
+			var err error
+			ignorePathFound, err = path.Match(pathIgnored, cleaned)
 			if err != nil {
 				http.Error(
 					w,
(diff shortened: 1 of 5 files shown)


pkg/filters/auth.go
```diff
@@ -107,17 +107,20 @@ func WithAuthorization(
...
-
 	return func(w http.ResponseWriter, req *http.Request) {
-		u, ok := request.UserFrom(req.Context())
-		if ok {
-			// Seemingly well-known headers to tell the upstream about user's identity
-			// so that the upstream can achieve the original goal of delegating RBAC authn/authz to kube-rbac-proxy
-			req.Header.Set(cfg.UserFieldName, u.GetName())
-			req.Header.Set(cfg.GroupsFieldName, strings.Join(u.GetGroups(), cfg.GroupSeparator))
+		req.Header.Del(cfg.UserFieldName)
+		req.Header.Del(cfg.GroupsFieldName)
```
ibihim, line 112: Why would we clean up the headers, if we are not supposed to set them?

cmd/kube-rbac-proxy/app/kube-rbac-proxy.go
```diff
@@ -283,9 +283,16 @@ func Run(cfg *completedProxyRunOptions) error {
 	}
 
 	handler := http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		cleaned := path.Clean(req.URL.Path)
+		if cleaned != req.URL.Path {
```
ibihim, line 287: 👍

pkg/filters/auth.go
yahlifried: When --auth-header-fields-enabled=false, the current code returns the handler unchanged, so client-supplied identity headers pass through to the upstream as-is. The proxy should be the only source of truth for these headers — if it's not setting them, no one should be. Stripping them when disabled is the complement of overwriting them when enabled.

ibihim: sorry for the ultra late response, I was on vacation and busy with other projects.

Well, this is an opinion :D ... with a regression (Hyrum's Law). This makes me worried merging such a thing.

ibihim: @stlaz for 2nd opinion. I am willing to get overturned on this one :stuck_out_tongue:

PR #447, opened 2026-07-15
Strip debug symbols

sdodson:
Follows DBG=1 pattern from kubernetes/kubernetes to produce an unstripped binary for debugging.

Reduces the binary size by approximately 32% when built with go1.26.5. 

Unstripped (DBG=1):
-rwxr-xr-x. 1 sdodson sdodson 77136200 Jul 15 13:19 kube-rbac-proxy-linux-amd64

Stripped:
-rwxr-xr-x. 1 sdodson sdodson 52355234 Jul 15 13:18 kube-rbac-proxy-linux-amd64

**Update:** addressed review feedback (https://github.com/kube-rbac-proxy/kube-rbac-proxy/pull/447/changes#r4027328789) by renaming the internal strip flags to `STRIP_LDFLAGS` so a user-supplied `GOLDFLAGS` is combined with them instead of being overwritten (e.g. `make build GOLDFLAGS="-v"` now works), and added `DBG_GCFLAGS` (`all=-N -l`) to disable optimizations/inlining when `DBG=1`, for a better debugging experience.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,16 @@ all: check-license build generate test
 GO111MODULE=on
 export GO111MODULE
 
+# Set DBG=1 to build without stripping debug symbols. Default produces a
+# stripped binary.
+DBG ?=
+
+ifeq ($(DBG),1)
+GOLDFLAGS ?=
+else
+GOLDFLAGS ?= -s -w
+endif
+
 PROGRAM_NAME?=kube-rbac-proxy
 # This repository moved to github.com/kube-rbac-proxy/kube-rbac-proxy.
 # Keep the old module path until the 1.0 release, where this is subject to change.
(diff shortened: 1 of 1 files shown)


sdodson:
@ibihim PTAL

Makefile
```diff
@@ -3,6 +3,16 @@ all: check-license build generate test
...
 
+# Set DBG=1 to build without stripping debug symbols. Default produces a
+# stripped binary.
+DBG ?=
+
+ifeq ($(DBG),1)
+GOLDFLAGS ?=
+else
+GOLDFLAGS ?= -s -w
+endif
```
ibihim, line 14: Nits:

```suggestion
ifeq ($(DBG),1)
DBG_GCFLAGS := all=-N -l
else
STRIP_LDFLAGS := -s -w
endif
```

So if someone defines:

```
make build GOLDFLAGS="-v"  # -> Use -v -s -w
```

And if someone wants to go the debugging route:

```
make build DBG=1 # -> Keep debug information, disable optimizations and inlining
```

WDYT?

ibihim:
Thank you for your contribution, it was an interesting to read more on this topic.

Makefile
sdodson: How's it look now?

s-urbaniak (approved):

PR #452, opened 2026-09-11
feat: add TLS curve preferences flag

damdo:
Following the addition of `--tls-curve-preferences` in [kubernetes/kubernetes](https://github.com/kubernetes/kubernetes/pull/137115), we add the equivalent flag to kube-rbac-proxy.

This lets operators restrict the TLS key-exchange mechanisms accepted by kube-rbac-proxy’s incoming secure listeners, supporting organizational TLS policies while retaining Go’s secure defaults when no preference is configured.

- Accepts comma-separated numeric Go `crypto/tls.CurveID` values.
- Uses Go’s internal preference order among the allowed groups.
- Preserves Go’s default curve selection when omitted.
- Rejects invalid, duplicate, or unsupported curve IDs at startup.
- Updates generated CLI help in the README.


cmd/kube-rbac-proxy/app/options/options_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+func TestTLSCurvePreferencesFlag(t *testing.T) {
+	o := NewProxyRunOptions()
+	flagSets := o.Flags()
+
+	if err := flagSets.FlagSet("kube-rbac-proxy").Parse([]string{"--tls-curve-preferences=23,29"}); err != nil {
+		t.Fatalf("parse flags: %v", err)
+	}
+
+	want := []int32{23, 29}
```
ibihim: Not sure if it is worth it:

```suggestion
	want := []int32{int32(tls.CurveP256), int32(tls.X25519)}
```

damdo: Done, thanks!

damdo:
The linting is failing already for unrelated reasons. I've raised another PR to fix that independently: https://github.com/kube-rbac-proxy/kube-rbac-proxy/pull/454

stlaz (approved):

PR #454, opened 2026-09-17
fix(lint): resolve TLS transport deprecation warnings

damdo:
Update deprecated networking APIs reported by the current `golangci-lint`
staticcheck rules.

- Replace `http2.Transport.DialTLS` with `DialTLSContext`, preserving h2c
  upstream behavior while allowing canceled requests to cancel the dial.
- Remove `net.Dialer.DualStack`, as Go enables fast fallback by default.

This restores clean full-repository linting with the current toolchain.

PR #455, opened 2026-09-17
*: update Go to 1.26.8 and dependencies, fix deprecated dialer options

ibihim:
## What

- **cmd/kube-rbac-proxy: replace deprecated upstream dialer options**
  - Drop `net.Dialer.DualStack`.
  - Replace `http2.Transport.DialTLS` with `DialTLSContext` in the `--upstream-force-h2c` transport.
  - Add `newUpstreamDialer`, which holds the dialer settings of `http.DefaultTransport` (30s dial timeout, 30s keep-alive), and use it for both upstream transports.
- **Go 1.26.5 → 1.26.8** in `go.mod` and the build and release workflows.
- **Dependency updates:**
  - k8s.io/* v0.36.2 → v0.36.4
  - golang.org/x/net v0.56.0 → v0.58.0
  - google.golang.org/grpc v1.80.0 → v1.83.2
  - github.com/google/cel-go v0.28.0 → v0.30.0
  - go.etcd.io/etcd v3.6.10 → v3.6.14
  - go.opentelemetry.io/otel v1.43.0 → v1.44.0
  - golang.org/x/crypto v0.53.0 → v0.57.0
  - other indirect dependencies

## Why

- **Lint is red on unchanged master.** CI runs golangci-lint `latest`, whose staticcheck now reports `DialTLS` and `DualStack` as deprecated (SA1019).
  - `DualStack` has been ignored since Fast Fallback became the default.
  - `DialTLSContext` gets a context derived from the proxied request, so the transport can cancel dials that are no longer needed.
- **The h2c upstream dial had no timeout.** It used `net.Dial`, which waits until the kernel gives up (about 2 min on Linux). It now uses the same dialer settings as the other upstream transports, so the values are defined once instead of being copied.
- **Keep Go and dependencies current.** This supersedes #449, #451 and #454.

## Behavior change

Only affects `--upstream-force-h2c`:
- Upstream dials time out after 30s.
- TCP keep-alive probes start after 30s idle instead of 15s.

## Testing

- `go build`, `go vet` and `make test-unit` (with `-race`) pass for each commit on Go 1.26.8.
- golangci-lint was not run locally, so this CI run is its first.


damdo (commented):
LGTM

s-urbaniak (approved):

